PR #6, opened 2019-05-14
Elaborate on how to raise issues to the steering committee

pmorie:
Follow-up from the May 10 2019 steering committee meeting.


isdal:
/lgtm
Thanks Paul for clarifying (and codifying) our decision from last week.

PR #16, opened 2019-06-10
Add guidelines for donated repositories

pmorie:
Per the June 7 2019 steering committee meeting.

### DIFF
--- a/REPOSITORY-GUIDELINES.md
+++ b/REPOSITORY-GUIDELINES.md
@@ -35,8 +35,14 @@ donated repositories must:
 
 - Adopt the Google CLA bot.
 - All contributors to the donated repository must have signed the Google CLA.
-- Boilerplate text across all files should attribute copyright as follows:
-  `"Copyright <year> The Knative Authors"
+- All code must adopt the standard Knative header, attributing copyright as follows:
+  `"Copyright <year> The Knative Authors"`
+ - Additions of the standard Knative header to code created by the contributors can 
+   occur post-transfer, but should ideally occur shortly thereafter.
+- An `AUTHORS` file should be created in the repo root, if one does not already exist, that 
+  includes the original authors of the code. 
+  See [knative/serving AUTHORS](https://github.com/knative/serving/blob/master/AUTHORS) for reference.
+  - Note that copyright notices should only be modified or removed by the people or organizations named in the notice.
 
 ## Removing Repositories
 
--- a/REPOSITORY-GUIDELINES.md
+++ b/REPOSITORY-GUIDELINES.md
@@ -20,12 +20,23 @@ Core repositories are all those located under the
 
 ### Core Repository Requirements
 
-- Repository must live under github.com/knative/project-name
-- Must adopt the Knative Code of Conduct
+- Repository must live under github.com/knative/project-name.
+- Must adopt the Knative Code of Conduct.
 - All code projects must use the Apache License version 2.0.
 - Documentation repositories must use Creative Commons License version 4.0.
 - All OWNERS must be members of the Knative community.
 - Repository creation must be approved by the Technical Oversight Committee.
+
+### Requirements for Repository Donation
+
+The Technical Oversight Committee may allow repositories to be donated to the
+Knative project. In *addition* to meeting the core repository requirements,
+donated repositories must:
+
+- Adopt the Google CLA bot.
+- All contributors to the donated repository must have signed the Google CLA.
+- Boilerplate text across all files should attribute copyright as follows:
+  `"Copyright <year> The Knative Authors"
 
 ## Removing Repositories
 


REPOSITORY-GUIDELINES.md
```diff
@@ -20,13 +20,24 @@ Core repositories are all those located under the
...
+### Requirements for Repository Donation
+
+The Technical Oversight Committee may allow repositories to be donated to the
+Knative project. In *addition* to meeting the core repository requirements,
+donated repositories must:
+
+- Adopt the Google CLA bot.
+- All contributors to the donated repository must have signed the Google CLA.
+- Boilerplate text across all files should attribute copyright as follows:
+  `"Copyright <year> The Knative Authors"
```
jchesterpivotal: I'm not a lawyer, but: would this replace existing copyright text, or would it augment existing copyright text?

```
    Copyright 1998-2019 ExampleCo International LLC 
```

Does that become

```
    Copyright 1998-2019 ExampleCo International LLC 
    Copyright 2019-current year The Knative Authors
```

Or does it ditch the `ExampleCo` attribution (assuming ExampleCo is willing and able to transfer copyright)?

csantanapr: I would vote to replace the copyright with the new one and avoid any shinanigans with dual copyright and copy left in the are litigations in the future.
Donating means donating everything including the copyright.

Attribution and credit, still there to individual contributors via git history/commits
Maybe clarify that the git repo should be donated with history in place.

REPOSITORY-GUIDELINES.md
```diff
@@ -20,13 +20,24 @@ Core repositories are all those located under the
...
+
```
csantanapr: Maybe Add bullets about some technical  guidelines:
- Donated repo should be moved with commit history using the Github move features that everything get's transferred only thing new is the org/user as the ownership and URL location. If repo is not in github then the a new repo is created and the master branch push to it including history.

REPOSITORY-GUIDELINES.md
rgregg: +1 to replacing the copyright, that's driven by the CLA as well -- the copyright is reassigned to Google / Knative authors. We shouldn't have dual copyright notices on the files.

If/when we can't have all contributors of the code sign the CLA, then we'll need to work through the details, but as long as we can meet that initial requirement this should be OK.

REPOSITORY-GUIDELINES.md
rgregg: I'm supportive of adding a guideline around bringing the full git repo history as part of the donation. Want to make sure we capture that as a "SHOULD" instead of a must however (as you have it written).

REPOSITORY-GUIDELINES.md
jcrossley3: All eight of the contributors have signed the CLA, fyi

jcrossley3: Re timing, is it ok to transfer first, and then add the copyright text, or must that happen prior to the transfer?

jchesterpivotal: I asked about this. Basically, it's not super important if it's one attribution or two, copyright arises from the particular facts of contributors and contributions rather than the copyright header. 

So I think that (1) we switch it to "The Knative Authors" and (2) we can make the change either before or after moving the repository, whichever's easiest.

rgregg: That's my understanding as well.

pmorie: So, is there a request to change the text here?

REPOSITORY-GUIDELINES.md
pmorie: Is there a request to change the text here?

rgregg:
I have a one or two ends to tie up still inside Google, and I'm working on getting closure so we can merge this and proceed. Unfortunately an entire team of folks I need to hear from are out of the office right now.

REPOSITORY-GUIDELINES.md
willnorris: Yes, do change the text to "The Knative Authors".

But two points to note:
1. copyright is not being assigned to anyone in this scenario.  Google's CLA is not a copyright assignment, it's a license (hence the name, Contributor License Agreement).  The original authors still retain copyright to the code they wrote.  When we say "The [Project] Authors", that still includes the original copyright holders, we're just using a broader term to identify them.

2. The reason we use this phrasing is for the benefit of downstream users who redistribute the code.  The Apache license requires that anyone who redistributes the code "retain [...] all copyright, patent, trademark, and attribution notices" (Section 4c).  If every file in the project had different copyright statements, then users would be obligated to find and reproduce all of them anytime they distribute the project.  We don't want them to have to do that, so we have simple, consistent copyright statements on all the files which makes it easier for them to fulfill their notice obligations.

pmorie:
> I have a one or two ends to tie up still inside Google, and I'm working on getting closure so we can merge this and proceed. Unfortunately an entire team of folks I need to hear from are out of the office right now.

Ack, I will avoid updating this PR until we get more info.

REPOSITORY-GUIDELINES.md
rgregg: Thanks Will. I've revised my comment above according. Sounds like we should be good to proceed with the current guidance proposed by @pmorie.

rgregg:
We should be good to go now Paul.

PR #27, opened 2019-08-15
Adjust requirements for Working Group leads to distinguish non-technical contributions

evankanderson:
Discussing with a number of WGs, there are a number of individuals who may take a critical non-coding role in both Working Groups and within their larger organization (hey look, that might sound like me!).

Rather than make "Working Group Lead" be "Approver++", make this a separate role whose primary purpose is to own the quality of outcomes produced by the working group through improving the decision-making process and organization of the working group.

### DIFF
--- a/ROLES.md
+++ b/ROLES.md
@@ -232,9 +232,16 @@ approver in an OWNERS file:
 
 ## Lead
 
-Working group leads, or just ‘leads’, are approvers of an entire area that have
-demonstrated good judgement and responsibility. Leads accept design proposals
-and approve design decisions for their area of ownership.
+Working group leads, or just ‘leads’, coordinate working group processes and
+decision-making. Leads are responsible for selecting appropriate processes and
+tracking and reporting on the working group's results to the TOC. Leads
+collaboratively take responsibilty for accepting design proposals and approving
+design decisions in their area of ownership.
+
+Working group leads may or may not overlap with code approvers -- the
+expectation is that working group leads are responsible for the overall quality
+of the working group's product (including documentation, testing, and
+usability), rather than just the service's code.
 
 ### Requirements
 
@@ -242,21 +249,27 @@ Getting to be a lead of an existing working group:
 
 - Recognized as having expertise in the group’s subject matter.
 
-- Approver for some part of the codebase for at least 3 months.
+- Substantial contributions to the working group. Note that these contributions
+  may include design documents, documentation, improvements to the working group
+  process, or other non-code work which has improved the working group's
+  deliverables.
 
-- Member for at least 1 year or 50% of project lifetime, whichever is shorter.
+- Member for at least 4 months.
 
-- Primary reviewer for 20 substantial PRs.
+- Sponsored by an existing Working Group lead or a majority of active working
+  group contributors.
 
-- Reviewed or merged at least 50 PRs.
+- Approved by the technical oversight committee.
 
-- Sponsored by the technical oversight committee.
+  - Nominations will be announced for at least a week prior to any decision;
+    concerns about the nomination may be flagged privately to any TOC member.
 
 Additional requirements for leads of a new working group:
 
-- Originally authored or contributed major functionality to the group's area.
+- Provided substantial leadership or functionality contributions in the group's
+  area.
 
-- An approver in the OWNERS file for the group’s code.
+- All proposed working group leads should be listed in the group's charter.
 
 ### Responsibilities and privileges
 
--- a/ROLES.md
+++ b/ROLES.md
@@ -233,10 +233,11 @@ approver in an OWNERS file:
 ## Lead
 
 Working group leads, or just ‘leads’, coordinate working group processes and
-decision-making. Leads are responsible for selecting appropriate processes and
-tracking and reporting on the working group's results to the TOC. Leads
-collaboratively take responsibilty for accepting design proposals and approving
-design decisions in their area of ownership.
+decision-making. A healthy working group has multiple leads, though not so many
+that decision making becomes difficult. Leads are responsible for selecting
+appropriate processes and tracking and reporting on the working group's results
+to the TOC. Leads collaboratively take responsibilty for accepting design
+proposals and approving design decisions in their area of ownership.
 
 Working group leads may or may not overlap with code approvers -- the
 expectation is that working group leads are responsible for the overall quality
--- a/ROLES.md
+++ b/ROLES.md
@@ -239,10 +239,16 @@ appropriate processes and tracking and reporting on the working group's results
 to the TOC. Leads collaboratively take responsibilty for accepting design
 proposals and approving design decisions in their area of ownership.
 
+Working group leads are responsible not only for the technical artifacts of
+their contribution area, but also the process and health of the community within
+that area. This manifests in well-understood and community-accepted decision
+processes and the growth of contributors in all roles, including mentoring
+additional leads.
+
 Working group leads may or may not overlap with code approvers -- the
 expectation is that working group leads are responsible for the overall quality
-of the working group's product (including documentation, testing, and
-usability), rather than just the service's code.
+of the working group's process and product (including documentation, testing,
+and usability), rather than just the service's code.
 
 ### Requirements
 


evankanderson:
/assign @pmorie @rgregg 
/hold

I believe this requires the Steering Committee to approve these changes.

evankanderson:
Ping

rgregg:
I'm supportive of these changes. 

This falls under the charter of the steering committee to approve this change. I'll add this to our discussion agenda for the next meeting.

ROLES.md
```diff
@@ -232,31 +232,44 @@ approver in an OWNERS file:
 
 ## Lead
 
-Working group leads, or just ‘leads’, are approvers of an entire area that have
-demonstrated good judgement and responsibility. Leads accept design proposals
-and approve design decisions for their area of ownership.
+Working group leads, or just ‘leads’, coordinate working group processes and
+decision-making. Leads are responsible for selecting appropriate processes and
```
duglin: replace the trailing ` and` with `,`

ROLES.md
```diff
@@ -232,31 +232,44 @@ approver in an OWNERS file:
...
+tracking and reporting on the working group's results to the TOC. Leads
+collaboratively take responsibilty for accepting design proposals and approving
```
duglin: This isn't clear on whether they're a mini-dictator or more like a moderator helping to group come to a consensus. I'd prefer the latter.

ROLES.md
```diff
@@ -232,31 +232,44 @@ approver in an OWNERS file:
...
 
-- Member for at least 1 year or 50% of project lifetime, whichever is shorter.
+- Member for at least 4 months.
 
-- Primary reviewer for 20 substantial PRs.
+- Sponsored by an existing Working Group lead or a majority of active working
+  group contributors.
 
-- Reviewed or merged at least 50 PRs.
+- Approved by the technical oversight committee.
```
duglin: We need to have a process by which a non-googler can be on the TOC, otherwise this can be read as "Approved by Google".

ROLES.md
evankanderson: I think this was intended to be:

> Leads are responsible for:
> * Selecting appropriate processes
> * Tracking and reporting on the working group's results to the TOC

Not:

> Leads are responsible for:
> * Selecting appropriate processes
> * Tracking
> * Reporting on the working group's results to the TOC.

ROLES.md
evankanderson: This is intended to suggest the latter (the word "collaboratively" is intended to suggest this). They are also responsible for delivering results of the process, so at some point there may come a "disagree and commit" decision that the leads (collectively) are responsible for.

Would adding a sentence like "A healthy working group has multiple leads, though not so many that decision making becomes difficult." help?

ROLES.md
evankanderson: I 100% agree that we need to describe the process for selecting TOC members.

I expect that this will be a longer discussion; the current `TECH-OVERSIGHT-COMMITTEE.md` file says:

> The members of the TOC are shown below. Membership in the TOC is determined by
> the [Steering committee](./STEERING-COMMITTEE.md).

But does not provide specific details about how this occurs.

evankanderson: The actual change here is:

    - - Sponsored by the technical oversight committee.
    + - Approved by the technical oversight committee.

(i.e. a change from "sponsored" to "approved" because that seemed to make the process more explicit)

evankanderson:
Hoping this can get approved or rejected/modified at the next Steering Committee Meeting.

evankanderson:
(Note that this needs an `/approve` from someone in this file: https://github.com/knative/community/blob/master/OWNERS_ALIASES)

ROLES.md
```diff
@@ -232,31 +232,45 @@ approver in an OWNERS file:
...
 ## Lead
 
-Working group leads, or just ‘leads’, are approvers of an entire area that have
-demonstrated good judgement and responsibility. Leads accept design proposals
-and approve design decisions for their area of ownership.
+Working group leads, or just ‘leads’, coordinate working group processes and
+decision-making. A healthy working group has multiple leads, though not so many
+that decision making becomes difficult. Leads are responsible for selecting
+appropriate processes and tracking and reporting on the working group's results
+to the TOC. Leads collaboratively take responsibilty for accepting design
```
duglin: This could be interpreted as the leads are the ones who approve/reject design proposals. What about shifting this slightly to talk about how it's their job to ensure design proposals get reviewed and approved/rejected via a community approval process? This switches them from being "mini dictators" into more like "moderators" or "facilitators". Given, in the next paragraph it talks about how they might not be code approvers, they may not necessarily have the technical background to take on the role of making these decisions. Plus, in an OSS project I think it should be the community who decides which direction to go, not one or two individuals. Each WG can then decide how they'll get consensus from the group - but an overall project-wide process would be nice too, to have consistency.

So, with that in mind, what about:
```
Leads are responsible for ensuring there is consistent and timely forward progress being
made within the working group. This includes making sure design proposals, issues and
PRs are reviewed and resolved by the working group by whatever community driven 
approval process is established.
```

evankanderson: I disagree -- I think the job of the WG leads (collectively) is to be **responsible for the outcomes and process**, not just facilitating reaching _any_ outcome. Being responsible for outcomes means that they should be empowered to set parameters or reject community decisions which violate WG or project principles. Being responsible for the process means that they need to be able to choose the process, within the general guidelines.

You bring up a good point that the WG leads are **also** responsible for the community health within the project. A well-respected WG lead may be able to take a tough position on one or two issues in the larger community, but repeatedly ignoring community input will result in a worse outcome even if the technical results might still be solid.

(Yes, this suggests that we should also look at a process for replacing WG leads where the WGs are not producing outcomes. I would welcome that, but it seems more critical at the moment to be able to expand our WG leadership to include more stakeholders from multiple viewpoints and employers than to winnow that number.)

duglin: If it means they’re dictators, which I’ve seen, then I disagree. This should be about community and everyone feeling that they have a voice. IMO having a set of maintainers where a certain number of them have to approve things, designs as well as code changes, is more community based (and has a clearer path to obtain the role) than a set of TOC approved leaders who may or may not be experts in the code. TBH I don’t think we need WG leads at all w.r.t. approving things, that’s what maintainers are for.  This is why I’d prefer for WG leads to focus more on administrivia goals like helping push towards the goal line, but not put a finger on the scale.

Abd4llA: IMHO one of the most effective aspects in Knative governance model is the clear setting of ownership (specifically technical ownership and decisions) which both encourages community discussions and decisions (WG calls, code reviewers, approvers, TOC, SC) while still giving the healthy balance and empowerment of WG leads, or TOC members. Even in K8s, there are such ownership such as [Tech Lead, Subproject Owner](https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-governance.md). I've seen how some communities fall into the temptations of "full consensus, community votes based decisions" where things fall apart when an opinion or a decision needs to be taken and the group is unable to move forward. 

As always, with great power comes great responsibility,  and for myself I've not seen so far this power in Knative being abused.  So I like this rephrasing and I'd strongly stay away from moving the WG Lead role to a moderation and facilitation role.

duglin: I wasn't saying "full consensus" (although that would be nice). We have maintainers, and they can LGTM or NOT LGTM - and we can decide how those votes are counted... (at least 2? 50+%?  whatever) that's all that's needed. This text could be interpreted as the lead can unilaterally make decisions and that's that part I'm questioning. They need to work "collaboratively" (as the text says) with the other maintainers to come to a resolution. But any implication that their vote carries more weight is the opposite of community driven. While I agree that we haven't had an issue in Kn yet, that's not really the point. When drafting up governance rules you write them for the worse case scenarios and then let people get a bit loose with them when things are going great. But if tension comes up, you then go back to "the law/rules" and that should force things back to a good state.

vagababov: Note that all  WGs have more than 1 and sometimes even 3 leads, which makes "unilateral" decision kind of hard (unless all of them agree, and in this case I might suggest that that _should_ carry some weight). Perhaps, requiring every WG to have more than one lead _is_ what we really need?

evankanderson:
/close

In favor of #56

PR #35, opened 2019-10-20
removes @mchmarny from list of steering committee members

mchmarny:
This PR removes @mchmarny from the list of steering committee members table


isdal:
Mark, since Jaice is taking your spot can you add him in the same PR?
Github handle is jdumars@ name is Jaice Singer DuMars

PR #37, opened 2019-10-25
Update membership and owners for Jaice

rgregg:


rgregg:
I think the issue was that Jaice hadn't accepted his organizational invite to the Knative org on GitHub. Should be fixed now.

PR #55, opened 2020-01-09
Remove unused roles

pmorie:
Per steering F2F

### DIFF
--- a/ROLES.md
+++ b/ROLES.md
@@ -37,17 +37,6 @@ table describes:
   </thead>
 
   <tr>
-    <td>Collaborator</td>
-    <td>Casual contributor to the project</td>
-    <td>Join [knative-users@](https://groups.google.com/forum/#!forum/knative-users) to get access to the team drive.</td>
-    <td>
-        <p>Can submit PRs</p>
-        <p>Commenting permission on the Knative Team drive</p>
-    </td>
-    <td>GitHub Organization</td>
-  </tr>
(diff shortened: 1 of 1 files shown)


isdal:
These are roles that we in practice don't enforce / use in the project.
/approve
/lgtm

ROLES.md
```diff
@@ -88,41 +77,8 @@ table describes:
     <td>Working Group</td>
   </tr>
 
-  <tr>
-    <td>Administrator</td>
```
mattmoor: This role exists today (we have knative-admins and #slack-admins).  Can we clarify what group populates this?

evankanderson: I think it's useful (and could be more useful over time) to have an explicit way to delegate parts of the TOC or SC work to people who are responsible for executing the decisions.

Examples would include TPMs or Community Managers who would have "Administrator" role to be able to manage Milestones, GitHub organizations, or Slack permissions without needing to interrupt the Steering Committee or TOC with those requests.

isdal: We wanted to clean up the roles where we haven't done a good job enforcing exactly how the role is granted or where we don't have anyone actually in the role. In practice the administration work has been done by SC members. If the workload is too high (and we can find someone interested in doing the work) we can create a new role for this.

Re: milestones and other work at the working group level: We have a second update coming which will define a role for that. Stay tuned.

PR #63, opened 2020-01-15
update web conf link for api-core

duglin:
/cc @dgerd 


dgerd:
/assign @pmorie 
/assign @rgregg 

whoops. realized that steering is the owner here. Can one of you approve?

PR #64, opened 2020-01-15
Create docs folder

samodell:
This moves the information about contributing to the docs to its own separate folder, allowing us to place a template page that had been living in the Docs repo here instead.

From what I understand, having a docs folder should also make it possible for us to be given permissions to approve changes within only the docs folder of the Community repo, so that we can update and maintain that information without needing someone from the Community approvers list to review.

Docs PR: https://github.com/knative/docs/pull/2106

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -6,3 +6,6 @@ aliases:
     - pmorie
     - rgregg
     - lindydonna
+  docs-team:
+    - samodell
+    - richieescarez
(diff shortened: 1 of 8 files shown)


samodell:
I thought the issue might have been that Jaice isn't a member of the Knative Org anymore -- https://github.com/orgs/knative/people?utf8=%E2%9C%93&query=jdumars

But that doesn't seem to have fixed the problem. @rgregg  any ideas?

samodell:
the problem was defining aliases in /docs folder instead of at the top level. Fixed it now!

PR #66, opened 2020-01-23
Point to the meeting minutes for the VC link

duglin:
Why?
1 - now there's only one spot to update when things change
2 - don't need a PR each time the VC info changes
3 - I keep clicking on the wrong link (the one here instead of the notes)

If people are ok with this I can do the same change for the WG doc too.


duglin:
ping @rgregg @pmorie @bsnchan

PR #105, opened 2020-03-05
Update the process for creating a new working group.

evankanderson:
Make it clear that community consensus should be clear before writing a charter.

Include example charters for reference.

/assign @mikehelmick 
/assign @mattmoor 
/assign @vaikas 

PTAL and suggest further refinements to the process.

### DIFF
--- a/mechanics/WORKING-GROUP-PROCESSES.md
+++ b/mechanics/WORKING-GROUP-PROCESSES.md
@@ -42,9 +42,16 @@ the number of working groups low in order to keep things manageable.
 
 ## Proposing a new working group
 
-If you’ve identified a substantial architectural area which would benefit from
-long-lived, concerted and focused design, then you should consider creating a
-new working group. To do so, you need to:
+Generally, working groups should be founded once sufficient community interest
+and discussion has occurred that there is general agreement that:
+
+1. The problem space is interesting and is worth long-term investment by Knative
+
+1. The problem is not well-covered by any of the existing chartered WGs
+
+Once the community has identified a substantial architectural area which would
+benefit from long-lived, concerted and focused design, then you should consider
+creating a new working group. To do so, you need to:
 
 - **Create a charter**. This should be a few paragraphs explaining:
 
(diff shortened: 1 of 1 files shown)


mechanics/WORKING-GROUP-PROCESSES.md
```diff
@@ -42,9 +42,18 @@ the number of working groups low in order to keep things manageable.
...
-new working group. To do so, you need to:
+Generally, working groups should be founded once sufficient community interest
+and discussion has occurred that there is general agreement that:
+
+1. The problem space is interesting and is worth long-term investment by Knative
+
+1. The problem is not well-covered by any of the existing chartered WGs
+
+The TOC and existing WGs should enable these conversations (e.g. setting up
+one-off meetings, creating slack channels, etc).  s˝Once the community has
```
mikehelmick: some stray characters here

mechanics/WORKING-GROUP-PROCESSES.md
```diff
@@ -42,9 +42,18 @@ the number of working groups low in order to keep things manageable.
 
 ## Proposing a new working group
 
-If you’ve identified a substantial architectural area which would benefit from
-long-lived, concerted and focused design, then you should consider creating a
-new working group. To do so, you need to:
+Generally, working groups should be founded once sufficient community interest
```
mikehelmick: Not sure if this is the right PR or the right doc - but I think it would be good to encourage this process over going off and solving something with the intent of trying to donate/merge it in later.

mechanics/WORKING-GROUP-PROCESSES.md
mattmoor:
```suggestion
one-off meetings, creating slack channels, etc).  Once the community has
```

mechanics/WORKING-GROUP-PROCESSES.md
```diff
@@ -42,9 +42,18 @@ the number of working groups low in order to keep things manageable.
...
+and discussion has occurred that there is general agreement that:
+
+1. The problem space is interesting and is worth long-term investment by Knative
+
+1. The problem is not well-covered by any of the existing chartered WGs
```
mattmoor: I think the way I'd put this is: "WG proposals shouldn't be a surprise to interested parts of the community".

By the time it reaches a WG proposal, it should be well socialized and the right parties should be pulled in to the formation.  I don't think we need consensus or a solution, but I think for inclusion upstream we want to be better about things having broad support from the community.

Donation (per Mike's comment above) should be limited to cases where there is early consensus on one particular approach.  I also think donation can happen after formation if that consensus arrives later as part of the WG process.

evankanderson: I added a paragraph warning against "code dumps" as the start of a WG:

> Note that "code dumps" are unlikely to be accepted as the beginning of a working
> group, though they can be useful to start a conversation. For most topics that
> meet the bar for a new WG, it's likely that other participants in the community
> have ideas and opinions, and up-front conversations can help to focus the shape
> and scope of the solution space.

mechanics/WORKING-GROUP-PROCESSES.md
evankanderson: I added a paragraph about this later.

evankanderson:
I totally failed to push my commits earlier. PTAL.

PR #108, opened 2020-03-24
k4leung4 steps down as Operations WG lead

k4leung4:


mattmoor:
Thanks for your service @k4leung4 😢 

/lgtm
/approve

PR #109, opened 2020-03-25
Update meeting schedule for client WG

rhuss:
The client working group decided to move to an alternating schedule
for the WG calls, so that people from APAC region can join, too.
For reasoning and the community poll, please see https://github.com/knative/client/issues/740
(and also the client WG call protocols).
So ecery odd-numbered week (like this week, which is calendar week 13),
the current timeslot is active, on every other week we want to try
out 3:30pm Central European time (CET or CEST)

Beside this PR, would it be possible to adapt the Knative calendar, too ?
Although we settled on an European timezone for the second timeslot,
as a good approximation we can also use 6:30am Pacific Time.


rhuss:
For the concrete calendar entries:

* 10:30am Pacific time, April 7th, 21th, May 5th, ...
* 6:30am Pacific time, March 31st, April 14th, April 28th, ...

PR #117, opened 2020-04-06
Fix the link to `DOCS-CONTRIBUTING.md`

bsideup:
it seems that the file got moved to the `docs/` folder but the link was not updated.


bsideup:
@googlebot I signed it!

PR #148, opened 2020-06-09
Reconcile ToC members in OWNER_ALIASES.

markusthoemmes:
as per title.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -10,6 +10,8 @@ aliases:
     - samodell
     - richieescarez
   technical-oversight-committee:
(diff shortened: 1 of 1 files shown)


OWNERS_ALIASES
```diff
@@ -10,6 +10,8 @@ aliases:
     - samodell
     - richieescarez
   technical-oversight-committee:
-    - vaikas
-    - mattmoor
     - evankanderson
+    - grantr
```
vagababov, line 14: if you're cleaning up we can probably remove rgregg and lindydonna as well...

PR #150, opened 2020-06-16
New Google representatives in steering

isdal:
Update steering committee membership with new Google representatives.


bsnchan:
Thanks for all the work you've done for Knative @isdal ! We're going to miss you.


/lgtm
/approve

PR #159, opened 2020-06-19
Add knative-prow-robot as an org admin

chizhg:
Add `knative-prow-robot` as an org admin for `knative` and `knative-sandbox`, so it has enough permissions to run `peribolos` jobs.

/assign @evankanderson

Please note this permission will have to be manually set for the first time.

### DIFF
--- a/peribolos/knative-sandbox.yaml
+++ b/peribolos/knative-sandbox.yaml
@@ -17,6 +17,7 @@ orgs:
     - googlebot
     - grantr
     - isdal
+    - knative-prow-robot
     - markusthoemmes
     - mattmoor
     - mbehrendt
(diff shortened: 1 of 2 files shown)


evankanderson:
I'm almost ready to do this for `knative-sandbox`. I'd like to wait a day or two to make sure nothing blows up before applying to `knative` (also to have time to do a check that permissions haven't drifted since they were submitted).

chizhg:
I only added the Prow job for `knative-sandbox`, see https://github.com/knative/test-infra/blob/master/config/prod/prow/jobs/custom/peribolos.yaml, so at this moment nothing will be done to `knative`.

But the admin permission is required for both `knative` and `knative-sandbox` org, so let's add it for both to also be in preparation for `knative`?

PR #170, opened 2020-06-25
Add discovery repo to peribolos

grantr:
Creating repo for request at https://github.com/knative/eventing/issues/3167


evankanderson:
(Sorry, missed this while I was on vacation.)

Should there be a WG which has write/admin permissions to allow them to manage projects, milestones, etc?

PR #172, opened 2020-06-29
Add Cris Zong to the Knative org

chizhg:
Cris is a new summer intern in Google, and will be working on Knative during these three months.

And by the way fix one syntax error in the `knative-sandbox.yaml` file as the key is duplicated.


chizhg:
/cc @chaodaiG 
FYI @yizong98

PR #185, opened 2020-07-14
Remove urasoko from peribolos config file to get rid of the job error

chizhg:
For unknown reason, the peribolos Prow job is failing when it processes `urasoko` as a member, and removing this person will resolve this error.

See https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-sandbox-peribolos/1283169654633664513 and https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-sandbox-peribolos-test/1283171920660926465.

/cc @mattmoor

### DIFF
--- a/peribolos/knative-sandbox.yaml
+++ b/peribolos/knative-sandbox.yaml
@@ -200,7 +200,6 @@ orgs:
     - troytop
     - trshafer
     - tzununbekov
-    - urasoko
     - uwefassnacht
     - vagababov
     - vaikas


mattmoor:
Is there an issue for tracking this down and adding them back that we can link?

cc @urasoko FYI

chizhg:
> Is there an issue for tracking this down and adding them back that we can link?
> 
> cc @urasoko FYI

Created https://github.com/knative/community/issues/186 for tracking.

PR #196, opened 2020-07-22
Add zhongduo to members list

zhongduo:
This pull request is to add zhongduo@google.com to the members list.

### DIFF
--- a/peribolos/knative.yaml
+++ b/peribolos/knative.yaml
@@ -484,6 +484,7 @@ orgs:
     - zhanggbj
     - zhangtbj
     - ZhiminXiang
+    - zhongduo
     - zhruan
     - zouyee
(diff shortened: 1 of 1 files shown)


mattmoor:
@evankanderson are we trying to keep the two orgs in sync at all?

evankanderson:
I don't think we can effectively do so; there are a large number of people in the `knative` org that never accepted invites to `knative-sandbox` (because Peribolos can only send invites, not actually add people to the org against their will), so I removed them from the `knative-sandbox` list.

I'm also not sure that there are any particular reasons (for example) to require that someone who wants to be an org member in `knative-sandbox` need also add themselves to `knative`, though I'd be willing to maintain that invariant.

/lgtm
/approve

PR #200, opened 2020-07-24
Document the process to sunset a feature from Knative.

### LINKED ISSUE #199: Document process to retire a user-facing feature
Autoscaling just did this successfully for custom-metrics.

Maybe Markus or Victor could write up the process that they used (3-4 paragraphs) and put it in `mechanics/`

markusthoemmes:
Fixes  #199.

This is roughly the process that we've followed in the Autoscaling WG to sunset the custom-metrics API.

### DIFF
--- a/mechanics/SUNSETTING-FEATURES.md
+++ b/mechanics/SUNSETTING-FEATURES.md
@@ -1,0 +1,61 @@
+---
+title: "Knative feature sunsetting"
+linkTitle: "Sunsetting features"
+weight: 50
+type: "docs"
+---
+
+# Knative Feature Sunsetting
+
+If a feature (especially in alpha stage) is getting no apparent usage and is
+creating mostly busy-work for the community to maintain, the respective working
+group should consider to sunset the respective feature. This needs to be in
+accordance to the
+[Knative release principles](RELEASE-VERSIONING-PRINCIPLES.md#api-support-principle)
+so the duration of the following process might vary depending on the state of
+the respective feature.
+
+## 1. Working Group decision
+
+First of all, there should be a decision at the working group level that there
+is a desire to remove the respective feature. This decision should also contain
+an explanation, stating why the feature can and should be dropped from Knative.
+
+## 2. Gauge usage
+
+While the working group itself might think there is no usage, that feeling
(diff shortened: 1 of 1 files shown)


mechanics/SUNSETTING-FEATURES.md
```diff
@@ -0,0 +1,61 @@
...
+
+## 3. Reevaluate the decision
+
+Once the data has been gathered, the working group should reconsider the
+decision with the collected data in mind.
+
+## 4. Sunset
+
+If the decision to remove the feature remains, it can now actually be removed.
+This might either be a deprecation notice for Beta and GA feature or a straight
```
vagababov, line 60: don't we need N releases for removing a GA feature since notice?

markusthoemmes: Yeah that's covered in the API principles linked above. Didn't want to repeat that here to avoid creating two places to update.

mechanics/SUNSETTING-FEATURES.md
```diff
@@ -0,0 +1,61 @@
...
+linkTitle: "Sunsetting features"
+weight: 50
+type: "docs"
+---
+
+# Knative Feature Sunsetting
+
+If a feature (especially in alpha stage) is getting no apparent usage and is
+creating mostly busy-work for the community to maintain, the respective working
+group should consider to sunset the respective feature. This needs to be in
```
evankanderson, line 12: I think there's a larger cost-benefit ratio to consider here. For example, even a beta API which turns out (for example) to have serious scalability implications (KService/cluster from 500 => 80, for example) should be sunsetted.

mechanics/SUNSETTING-FEATURES.md
evankanderson, line 60: Not sure if we want to include appropriate feedback (e.g. retiring an API-level knob => add an API-level condition if the retired feature is used) as part of the deprecation.

abrennan89:
See comment https://github.com/knative/community/issues/199#issuecomment-665157145 re docs process

markusthoemmes:
@evankanderson @abrennan89 Incorporated your feedback.

Issue #166, opened 2020-06-25
State of Observability


mattmoor:
Rotating the WG update schedule, Observability would have been next but was "struck through", and the intent of this was not completely clear to me.  It was not my understanding that we had closed on the topic of "what to do with Observability" (it is still in our community docs).

I think that we either need to get Observability back into the "WG update" rotation or archive the WG and clean up the references in community and elsewhere.

/assign @mdemirhan @evankanderson @mattmoor

mattmoor:
I'm going to put this on this week's agenda since this came up in slack: https://knative.slack.com/archives/C964EFQAW/p1594674091070300?thread_ts=1594388710.067600&cid=C964EFQAW

evankanderson:
We care about making it possible to observe the working of Knative components and make it easy to integrate with o11y backends. Individual WGs will still be responsible for instrumenting their components for metrics, tracing, etc as needed, but we don't have anyone currently with bandwidth to drive larger improvements to observability for workloads running on Knative.

I propose that we [dissolve this WG](https://github.com/knative/community/blob/master/mechanics/WORKING-GROUP-PROCESSES.md#dissolving-a-working-group) in 3 weeks (6 Aug 2020) unless multiple contributors step up.

Note that retiring this WG does not prevent it from being re-charted in the future.

mattmoor:
Talked about this today at TOC.  We'll move ahead with retiring the Working Group.

We can always form a new working group given sufficient activation energy. 😉

PR #207, opened 2020-08-06
Retire the Observability Working Group

mattmoor:
Fixes: https://github.com/knative/community/issues/166

### DIFF
--- a/working-groups/WORKING-GROUPS.md
+++ b/working-groups/WORKING-GROUPS.md
@@ -58,7 +58,6 @@ The current working groups are:
 - [Event Delivery](#event-delivery)
 - [Eventing Sources](#eventing-sources)
 - [Networking](#networking)
-- [Observability](#observability)
 - [Operations](#operations)
 - [Productivity](#productivity)
 - [Scaling](#scaling)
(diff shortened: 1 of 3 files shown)


working-groups/WORKING-GROUPS.md
```diff
@@ -201,22 +200,6 @@ interest include: load balancing, routing, DNS configuration and TLS support.
 | <img width="30px" src="https://github.com/tcnghia.png">   | Nghia Tran       | Google  | [tcnghia](https://github.com/tcnghia)     |
 | <img width="30px" src="https://github.com/mdemirhan.png"> | Mustafa Demirhan | Google  | [mdemirhan](https://github.com/mdemirhan) |
 
-## Observability
```
grantr, line 204: I think it would be useful to archive the WG details in a separate file so the links to meeting notes, docs, and slack channel are preserved without having to dig through the team drive. We can also add an epitaph that explains when and why the WG was archived. This would be useful context for whoever tries to spin this WG up again (mentioned as a possibility in the TOC meeting today)

evankanderson: I think Max is actually interested in a "performance" WG, which would have a different focus from "observability" in that they would be looking primarily at how to improve measurable system qualities, rather than providing developers insight when their applications on the platform are failing.

(You could send a PR to Matt's branch for this.)

mattmoor: We didn't do this for Build, and I'm not super worried about recovering context (given commit history).

A graveyard is just another thing to maintain, so I'm a bit reluctant to take that on.

Issue #33, opened 2019-09-27
Clarify formula for membership of Steering committee


sebgoa:
Currently the SC is populated based on:

> Seats on the steering committee are allocated based upon contribution to the project by an organization. No final decision has been made on the exact formula.

I think it is time to clarify this formula. Until the formula is clearly spelled out this does not feel like an open governance to me and with Google's majority of 4 seats, this feels too much like a Google project. This is starting to cause a perception issue in the large k8s/cloud-native community which I strongly believe needs to be addressed.

I would recommend to look at the way kubernetes setup the bootstrap committee and subsequently held elections.

rgregg:
I've personally struggled with how we can fairly measure contribution to the project. Not all contributions come in the form of measurable data. There's certainly things like LOC, PRs, issues opened, issues resolved, issues worked on, leadership, presentations, code reviews, $'s spent, etc. Trying to find an exact formula that we can plug data into and produce a meaningful output hasn't been a priority, although maybe it should be. 

I'd be interested to know if there are examples we can reference or other formulas that have been used in other projects to measure the contributions of the community. I'd rather not invent our own formula for Knative if we can avoid it.

In the past we've referenced the [CNCF DevStats page for Knative](https://knative.teststats.cncf.io/d/5/companies-table?orgId=1), to gather some insights into contributions to the project, although it feels like a bit of a black box in terms of what that measures. If we used that metric alone, the SC make-up would be even more skewed, so we originally balanced it based on orders of magnitude of contributions across the companies listed.

I'm open to suggestions for how we can measure contributions objectively to create a formula, but I don't see us reconsidering the organization style membership of the Steering committee right now.

bsnchan:
Conversation and Discussions happening here: https://docs.google.com/document/d/1wkQ0okNxdrno88Wj-J7uqPIPMB7Uvge_kuV4RXRI4JE/edit#

PR #214, opened 2020-08-12
WIP: proposal for steering composition mechanics

pmorie:
This PR contains a WIP proposal for a new steering committee charter, which is made not only by myself but also by Brenda Chan and Michael Behrendt. For clarity, this proposal is not endorsed by the steering committee at the time of posting, and is not made on behalf of the steering committee.

The purpose of this proposal is to foster discussion of the mechanics for how steering is formed and maintained in a way that upholds the project [values](https://github.com/knative/community/blob/master/VALUES.md) of visibility, transparency, and feedback from our community.

We welcome your comments on this proposal and invite you to join us for a discussion on this topic in the August 19 public meeting of the steering committee.

If you would prefer to share your thoughts anonymously, you can use [this form](https://forms.gle/NgxsT7ojfRRbdPBVA). Note: responses to this form will be anonymous, but public.

Survey results can be accessed [here](https://docs.google.com/spreadsheets/d/1Lpg8vEvpM_wz0_KSqPHBBCq64lyT0tX7Vg5EkrSL6r8/edit?usp=sharing).

See https://github.com/knative/community/issues/33

### DIFF
--- a/STEERING-COMMITTEE.md
+++ b/STEERING-COMMITTEE.md
@@ -212,7 +212,7 @@ representation discussed below. This is the same process used for TOC elections.
 ### Limitations on Company Representation
 
 No more than two seats may be held by employees of the same organization (or
-consortium, in the case of companies owning each other).  Additionally, the two
+conglomerate, in the case of companies owning each other).  Additionally, the two
 End-User seats may not belong to employees of the same organization. If the
 results of an election result in greater than two employees of the same
 organization, the lowest vote getters from any particular employer will be
--- a/STEERING-COMMITTEE.md
+++ b/STEERING-COMMITTEE.md
@@ -38,7 +38,6 @@ the way we run this committee, based on feedback from the community.
    disposition of those resources.
 1. Manage the Knative brand and decide which things can be called "Knative" and
    how that mark can be used in relation to other efforts or vendors.
-1. Confirm/reject nominations to the Technical Oversight Committee.
 1. Receive and handle reports about [code of conduct](./CODE-OF-CONDUCT.md)
    violations and maintain confidentiality.
 1. Receive security reports; work with the appropriate technical leads to accept
@@ -189,7 +188,8 @@ requirements based on community feedback.
 Community members must be eligible to vote in order to stand for election (this
 includes voters who qualify for an exception). Candidates may self-nominate or
 be nominated by another eligible member. There are no term limits for KSC
-members.
+members. Nothing prevents a qualified member from serving on both the TOC and SC
+simultaneously.
 
 To run for a Contributing seat, a candidate must additionally be at least a
 project Member as defined in [ROLES.md](https://github.com/knative/community/blob/master/ROLES.md).
@@ -261,6 +261,18 @@ Amendments are accepted with majority consent of the committee.
 
 Proposals and amendments to the charter are available for at least a period of
 one week for comments and questions before a vote will occur.
+
+## Authority, Facilitation, and Decision Making in Knative
+
+Ideally most decisions will be made at the lowest possible level within the
+project: within individual working groups. When this is not possible for
+technical decisions, the TOC can help facilitate a conversation to work through
+the contended issue. When facilitation by the TOC does not resolve the contended
+issue, the Steering Committee can help facilitate. When facilitation by the SC
+does not resolve the contention, the SC may have to make a decision.
+
+Note that if the SC is called to resolve contended decisions regularly, it is a
+symptom of a larger in the community that will need to be addressed.
 
 ---
 
--- a/STEERING-COMMITTEE.md
+++ b/STEERING-COMMITTEE.md
@@ -19,10 +19,11 @@ the way we run this committee, based on feedback from the community.
 - [Committee Meetings](#committee-meetings)
 - [Committee Mechanics](#committee-mechanics)
 - [Committee Members](#committee-members)
-  - [Allocation of seats](#allocation-of-seats)
 - [Decision process](#decision-process)
+- [Raising issues for the SC](#raising-issue-to-the-steering-committee)
+- [Elections](#election-procedure)
+- [Vacancies](#vacancies)
 - [Changes to the charter](#changes-to-the-charter)
-- [Getting in touch](#getting-in-touch)
 
 ## Charter
 
@@ -37,8 +38,6 @@ the way we run this committee, based on feedback from the community.
    disposition of those resources.
 1. Manage the Knative brand and decide which things can be called "Knative" and
    how that mark can be used in relation to other efforts or vendors.
-1. Confirm/reject nominations to the KSC from organizations who are allocated
-   seats.
 1. Confirm/reject nominations to the Technical Oversight Committee.
 1. Receive and handle reports about [code of conduct](./CODE-OF-CONDUCT.md)
    violations and maintain confidentiality.
@@ -60,28 +59,26 @@ committee currently recognizes this delegated authority for:
 
 KSC meets every two weeks, or as-needed. Meetings are held online.
 
-Given the private nature of many of these discussions (e.g. privacy, private
+Given the private nature of some of these discussions (e.g. privacy, private
 emails to the committee, code of conduct violations, escalations, disputes
-between members, security reports, etc.) meetings are held in private.
+between members, security reports, etc.) some meetings are held in private.
 
-Meeting notes are available to members of the knative-dev mailing list (link to
-be added).
+Meeting notes are available to members of the [knative-dev mailing list](https://groups.google.com/forum/#!forum/knative-dev).
 
 Questions and proposals for changes to governance are posted as
 [issues in the docs repo](https://github.com/knative/docs/issues), and the KSC
-invites your feedback there. See [Getting in touch](#getting-in-touch) for other
+invites your feedback there. See [Raising Issues](#raising-issue-to-the-steering-committee) for other
 options.
 
 ## Committee members
 
-Seats on the Steering Committee are held by an organization, not by the
-individual.
+Seats on the Steering Committee are held by an individual, not by their employer.
 
-The committee was created as the project was in its infancy, in order to tackle
-governance and overall project strategy. Because of the nature of the project
-and funding required, it was decided that strong corporate leadership was
-necessary for the project to ensure velocity. As the project grows and matures
-the KSC will, from time to time, consider if this policy should be changed.
+When the Knative project was created, it was assigned strong corporate
+leadership through organizationally appointed steering committee members. As a
+mature project with production adoption, the project has moved to a democratic
+process of selecting new SC members.  This process may change again in the
+future as Knative continues to grow.
 
 The current membership of the committee is currently (listed alphabetically by
 first name):
@@ -97,39 +94,8 @@ first name):
 |  | Open \*      | Google       |  |
 
 
-\* Open seats have been allocated to the organization per the committee rules,
-  but have not been assigned to a member of the organization.
-
-### Allocation of seats
-
-Seats on the steering committee are allocated based upon contribution to the
-project by an organization. No final decision has been made on the exact
-formula.
-
-As the project continues to grow, we expect to add additional seats to the
-committee, ensuring that those contributing to the project are properly
-represented.
-
-- After a seat is allocated to an organization, the organization shall nominate
-  a candidate to be confirmed by KSC. The committee reserves the right to not
-  confirm a candidate, in which the organization would need to nominate a new
-  candidate
-- Members of the committee may step down at any time. When a member steps down,
-  their organization shall nominate a new candidate.
-- If a member leaves their organization, the organization must nominate a new
-  committee member to replace them following the nomination and confirmation
-  process.
-- If an organization is unable to seat a candidate, the KSC reserves the right
-  to reallocate the seat to another organization.
-- Changes to the number of seats, which company seats are allocated to, and
-  nominations to the committee are confirmed by majority vote of the committee
-  members.
-- In situations where the organization which holds a seat is no longer a viable
-  entity (e.g. merger, dissolution, bankruptcy) the KSC will make a decision on
-  how to reallocate that seat. Seats do not automatically transfer to any
-  organization.
-- Members on the committee have a 1 year term from their confirmation, and may
-  be reconfirmed to the committee at the end of their term.
+\* Open seats have been allocated to the organization per the prior committee rules,
+  but have not been assigned to a member of the organization or replaced with elected seats.
 
 ## Decision process
 
@@ -143,14 +109,6 @@ Quorum is considered reached when at least half of the members are present.
 In case of extended absence, the organization of the absent member may appoint a
 single delegate from the same company during the absence.
 
-## Changes to the charter
-
-Changes to the KSC charter may be proposed via a Pull Request on the charter
-itself. Amendments are accepted with majority consent of the committee.
-
-Proposals and amendments to the charter are available for at least a period of
-one week for comments and questions before a vote will occur.
-
 ## Raising issues to the steering committee
 
 There are two ways to raise issues to the steering committee for decision:
@@ -162,11 +120,155 @@ There are two ways to raise issues to the steering committee for decision:
 2. Open an issue on a knative repository and indicate that you would like
    attention from the steering committee.
 
+## Composition
+
+The steering committee will have 7 seats.  Five of the seats will be
+Contributing seats, open to any project contributor.  Two of the seats will be
+End-User seats, open only to active end-user community members.  See Candidate
+Eligibility for definitions of these categories.
+
+Steering Committee members will serve for 2-year terms, staggered in order to
+preserve continuity.  Every year either two or three contributor seats will be
+elected, and one end user seat will be elected.
+
+## Election Procedure
+
+### Timeline
+
+Steering Committee elections shall be held in October or November of each year.
+Six weeks or more before the election, the Steering Committee shall appoint
+Election Officer(s) (see below).  Four weeks or more before the election, the
+Election Officer(s) will issue a call for nominations, publish the list of
+voters, and open the call for exceptions.  One week before the election the call
+for nominations and exceptions will be closed.  The election will be open for
+voting not less than two weeks and not more than four.  The results of the
+election shall be announced within one week of closing the election.  New
+Steering Committee members will take office on December 1 of each year.
+
+### Election Officer(s)
+
+Six weeks or more before the election, the Steering Committee will appoint
+between one and three Election Officer(s) to administer the election.  Elections
+Officers will be Knative community members in good standing who are eligible to
+vote, are not running for Steering in that election, and can make a public
+promise of impartiality.  They will be responsible for:
+
+* Making all announcements associated with the election
+* Preparing and distributing electronic ballots
+* Judging exception requests
+* Assisting candidates in preparing and sharing statements
+* Tallying voting results according to the rules in this charter
+
+### Eligibility to Vote
+
+Anyone who has at least 50 contributions in the last 12 months is eligible to
+vote in the Steering election. Contributions are defined as opening PRs,
+reviewing and commenting on PRs, opening and commenting on issues, writing
+design docs, commenting on design docs, helping people on slack, participating
+in working groups, and other efforts that help advance Knative the project.
+
+[This
+dashboard]((https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year)
+shows only GitHub based contributions and does not capture all the contributions
+we value. We expect this metric not to capture everyone who should be eligible
+to vote. If a community member has had significant contributions over the past
+year but is not captured in the [stats.knative.dev
+dashboard](https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year),
+they will be able to submit an exception form to the Elections Officer(s) who
+will then review and determine whether this member should be marked as an
+exception.  All exceptions, and the reasons for them, will be recorded in a log
+that will be available to Steering and the TOC.
+
+All eligible voters will be captured at
+knative/community/steering-elections/$YEAR/voters.md and the voters’ guide will
+be captured at knative/community/steering-elections/$YEAR/README.md, similar to
+the kubernetes election process and identical to the TOC election process.
+
+We are committed to an inclusive process and will adapt future eligibility
+requirements based on community feedback.
+
+### Candidate Eligibility
+
+Community members must be eligible to vote in order to stand for election (this
+includes voters who qualify for an exception). Candidates may self-nominate or
+be nominated by another eligible member. The approximate time commitment of a
+Steering member is around 5 hours per week.
+
+To run for a Contributing seat, a candidate must additionally be at least a
+project Member as defined in Roles.md.
+
+To run for an End-User seat, a candidate must be employed by, or own, an
+organization that does not sell products or services derived from Knative.
+Staff with in-house Knative deployments, as well as application developers who
+create applications that run on Knative, are eligible (and encouraged to run)
+for these seats.  Additionally, the candidate must participate in the
+knative-users mailing list and/or community-facing Knative Slack channels.
+
+### Voting Procedure
+
+Elections will be held using a time-limited
+[Condorcet](https://en.wikipedia.org/wiki/Condorcet_method) ranking on
+[CIVS](http://civs.cs.cornell.edu/) using the Schulze method. The top
+vote-getters will be elected to the open seats, with the exceptions for company
+representation discussed below. This is the same process used for TOC elections.
+
+### Limitations on Company Representation
+
+No more than two seats may be held by employees of the same organization (or
+consortium, in the case of companies owning each other).  Additionally, the two
+End-User seats may not belong to employees of the same organization. If the
+results of an election result in greater than two employees of the same
+organization, the lowest vote getters from any particular employer will be
+removed until representation on the committee is down to two.
+
+If employers change because of job changes, acquisitions, or other events, in a
+way that would yield more than 2 seats being held by employees of the same
+organization, sufficient members of the committee must resign until only two
+employees of the same employer are left. If it is impossible to find sufficient
+members to resign, all employees of that organization will be removed and new
+special elections held. In the event of a question of company membership (for
+example evaluating independence of corporate subsidiaries) a majority of all
+non-involved Steering Committee members will decide.
+
+## Vacancies
+
+In the event of a resignation or other loss of an elected SC member, the
+candidate with the next most votes from the previous election will be offered
+the seat, provided that person otherwise qualifies to join the SC. This process
+will continue until the seat is filled.
+
+In case this fails to fill the seat, a special election for that position will
+be held as soon as possible, unless the regular SC election is less than 7 weeks
+away.  Eligible voters from the most recent election will vote in the special
+election (ie: eligibility will not be redetermined at the time of the special
+election). Any replacement SC member will serve out the remainder of the term
+for the person they are replacing, regardless of the length of that remainder.
+
+## Transition
+
+In order to transition from the original, organization-appointed Steering
+Committee structure, elections will be phased in over 14 months.
+
+In October 2020, four positions will be elected, including three Contributing
+seats and one End-User seat.  One Google, one Red Hat/IBM, and one VMWare seat
+will be retained from the current Steering Committee.
+
+In October or November 2021, the three "old SC" seats will be replaced by
+elected seats (two Contributing and one End-User).
+
+## Changes to the charter
+
+Changes to the KSC charter may be proposed via a Pull Request on the charter
+itself. Amendments are accepted with majority consent of the committee.
+
+Proposals and amendments to the charter are available for at least a period of
+one week for comments and questions before a vote will occur.
+
 ---
 
 Portions of this document are adapted from the
 [Istio Steering Committee](https://github.com/istio/community/blob/master/STEERING-COMMITTEE.md)
-documentation, which is licensed under the Apache License 2.0.
+documentation, and other portions are copied from the [Kubernetes Steering Committee](https://github.com/kubernetes/steering/blob/master/elections.md), both of which are licensed under the Apache License 2.0.
 
 Except as otherwise noted, the content of this page is licensed under the
 [Creative Commons Attribution 4.0 License](https://creativecommons.org/licenses/by/4.0/),
--- a/STEERING-COMMITTEE.md
+++ b/STEERING-COMMITTEE.md
@@ -64,22 +64,18 @@ emails to the committee, code of conduct violations, escalations, disputes
 between members, security reports, etc.) some meetings are held in private.
 
 Meeting notes are available to members of the [knative-dev mailing
-list](https://groups.google.com/forum/#!forum/knative-dev). Public meetings will
+list](https://groups.google.com/forum/#!forum/knative-dev) (note: these are
+accessible to members of the knative-users list as well). Public meetings will
 be recorded and the recordings made available publicly.
 
-Questions and proposals for changes to governance are posted as
-[issues in the docs repo](https://github.com/knative/docs/issues), and the KSC
-invites your feedback there. See [Raising Issues](#raising-issue-to-the-steering-committee) for other
-options.
+Questions and proposals for changes to governance are posted as [issues in the
+community repo](https://github.com/knative/community/issues), and the KSC
+invites your feedback there. See [Raising
+Issues](#raising-issue-to-the-steering-committee) for other options.
 
 ## Committee members
 
 Seats on the Steering Committee are held by an individual, not by their employer.
-
-When the Knative project was created, it was assigned strong corporate
-leadership through organizationally appointed steering committee members. As a
-mature project with production adoption, the project has moved to a democratic
-process of electing new KSC members.
 
 The current membership of the committee is currently (listed alphabetically by
 first name):
@@ -103,7 +99,7 @@ first name):
 The steering committee desires to always reach consensus.
 
 Decisions are made in meetings when a quorum of the members are present and may
-pass with majority of the committee supporting it.
+pass with at least half the members of the committee supporting it.
 
 Quorum is considered reached when at least half of the members are present.
 
@@ -120,10 +116,10 @@ There are two ways to raise issues to the steering committee for decision:
 
 ## Composition
 
-The steering committee has 7 seats.  Five of the seats will be Contributing
-seats, open to any project contributor.  Two of the seats will be End-User
-seats, open only to active end-user community members.  See Candidate
-Eligibility for definitions of these categories.
+The steering committee has 7 seats.  Five of the seats are Contributing
+seats, open to any project contributor.  Two of the seats are End-User
+seats, open only to active end-user community members.  See [candidate
+eligibility](#candidate-eligibility) for definitions of these categories.
 
 Steering Committee members serve for 2-year terms, staggered in order to
 preserve continuity.  Every year either two or three contributor seats are
@@ -148,8 +144,9 @@ members will take office on December 1 of each year.
 Six weeks or more before the election, the Steering Committee will appoint
 between one and three Election Officer(s) to administer the election.  Elections
 Officers will be Knative community members in good standing who are eligible to
-vote, are not running for Steering in that election, and can make a public
-promise of impartiality.  They will be responsible for:
+vote, are not running for Steering in that election, who are not currently part
+of the Steering Committee and can make a public promise of impartiality.  They
+will be responsible for:
 
 * Making all announcements associated with the election
 * Preparing and distributing electronic ballots
@@ -166,21 +163,23 @@ design docs, commenting on design docs, helping people on slack, participating
 in working groups, and other efforts that help advance Knative the project.
 
 [This
-dashboard]((https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year)
+dashboard](https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year)
 shows only GitHub based contributions and does not capture all the contributions
 we value. We expect this metric not to capture everyone who should be eligible
 to vote. If a community member has had significant contributions over the past
 year but is not captured in the [stats.knative.dev
 dashboard](https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year),
 they will be able to submit an exception form to the Elections Officer(s) who
-will then review and determine whether this member should be marked as an
-exception.  All exceptions, and the reasons for them, will be recorded in a log
-that will be available to Steering and the TOC.
+will then review and determine whether this member should be eligible to vote.
+All exceptions, and the reasons for them, will be recorded in a log that will be
+available to Steering and the TOC.
 
 The electoral roll of all eligible voters will be captured at
 knative/community/steering-elections/$YEAR/voters.md and the voters’ guide will
 be captured at knative/community/steering-elections/$YEAR/README.md, similar to
-the kubernetes election process and identical to the TOC election process.
+the [kubernetes election
+process](https://github.com/kubernetes/steering/blob/master/elections.md) and
+identical to the TOC election process.
 
 We are committed to an inclusive process and will adapt future eligibility
 requirements based on community feedback.
@@ -193,7 +192,7 @@ be nominated by another eligible member. There are no term limits for KSC
 members.
 
 To run for a Contributing seat, a candidate must additionally be at least a
-project Member as defined in Roles.md.
+project Member as defined in [ROLES.md](https://github.com/knative/community/blob/master/ROLES.md).
 
 To run for an End-User seat, a candidate must be employed by an organization
 that does not sell products or services derived from Knative. Staff with
--- a/STEERING-COMMITTEE.md
+++ b/STEERING-COMMITTEE.md
@@ -63,7 +63,9 @@ Given the private nature of some of these discussions (e.g. privacy, private
 emails to the committee, code of conduct violations, escalations, disputes
 between members, security reports, etc.) some meetings are held in private.
 
-Meeting notes are available to members of the [knative-dev mailing list](https://groups.google.com/forum/#!forum/knative-dev).
+Meeting notes are available to members of the [knative-dev mailing
+list](https://groups.google.com/forum/#!forum/knative-dev). Public meetings will
+be recorded and the recordings made available publicly.
 
 Questions and proposals for changes to governance are posted as
 [issues in the docs repo](https://github.com/knative/docs/issues), and the KSC
@@ -77,8 +79,7 @@ Seats on the Steering Committee are held by an individual, not by their employer
 When the Knative project was created, it was assigned strong corporate
 leadership through organizationally appointed steering committee members. As a
 mature project with production adoption, the project has moved to a democratic
-process of selecting new SC members.  This process may change again in the
-future as Knative continues to grow.
+process of electing new KSC members.
 
 The current membership of the committee is currently (listed alphabetically by
 first name):
@@ -119,28 +120,28 @@ There are two ways to raise issues to the steering committee for decision:
 
 ## Composition
 
-The steering committee will have 7 seats.  Five of the seats will be
-Contributing seats, open to any project contributor.  Two of the seats will be
-End-User seats, open only to active end-user community members.  See Candidate
+The steering committee has 7 seats.  Five of the seats will be Contributing
+seats, open to any project contributor.  Two of the seats will be End-User
+seats, open only to active end-user community members.  See Candidate
 Eligibility for definitions of these categories.
 
-Steering Committee members will serve for 2-year terms, staggered in order to
-preserve continuity.  Every year either two or three contributor seats will be
-elected, and one end user seat will be elected.
+Steering Committee members serve for 2-year terms, staggered in order to
+preserve continuity.  Every year either two or three contributor seats are
+elected, and one end user seat is elected.
 
 ## Election Procedure
 
 ### Timeline
 
-Steering Committee elections shall be held in October or November of each year.
-Six weeks or more before the election, the Steering Committee shall appoint
-Election Officer(s) (see below).  Four weeks or more before the election, the
-Election Officer(s) will issue a call for nominations, publish the list of
-voters, and open the call for exceptions.  One week before the election the call
-for nominations and exceptions will be closed.  The election will be open for
-voting not less than two weeks and not more than four.  The results of the
-election shall be announced within one week of closing the election.  New
-Steering Committee members will take office on December 1 of each year.
+Steering Committee elections are held annually. Six weeks or more before the
+election, the Steering Committee will appoint Election Officer(s) (see below).
+Four weeks or more before the election, the Election Officer(s) will issue a
+call for nominations, publish the list of voters, and open the call for
+exceptions.  One week before the election the call for nominations and
+exceptions will be closed.  The election will be open for voting not less than
+two weeks and not more than four.  The results of the election will be
+announced within one week of closing the election.  New Steering Committee
+members will take office on December 1 of each year.
 
 ### Election Officer(s)
 
@@ -176,7 +177,7 @@ will then review and determine whether this member should be marked as an
 exception.  All exceptions, and the reasons for them, will be recorded in a log
 that will be available to Steering and the TOC.
 
-All eligible voters will be captured at
+The electoral roll of all eligible voters will be captured at
 knative/community/steering-elections/$YEAR/voters.md and the voters’ guide will
 be captured at knative/community/steering-elections/$YEAR/README.md, similar to
 the kubernetes election process and identical to the TOC election process.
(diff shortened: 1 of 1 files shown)


duglin:
@pmorie thanks for this. Question: give the list of SC members are more technical in nature than today's SC members (not implying anything :-)  other than today SC folks are not necessarily required to be contributors), I think this makes the line between the SC and TOC kind of blurry and it makes me wonder why we'd need two layers within a project of this size. IOW, can we just merge the two groups into one?

bsnchan:
For some context, these are guiding principles that we aligned to when we wrote up this charter:

**Guiding Principles**

* The steering committee seat should be earned by individuals and not company participation
* Members of the steering committee should be exemplary members of the community who have demonstrated our community values
* The steering committee should be made up of individuals from different vendors and customers so that we have a diverse group of thought and input
* Limits on company representation should prevent any single vendor from having unilateral control over steering of the project

STEERING-COMMITTEE.md
```diff
@@ -143,14 +109,6 @@ Quorum is considered reached when at least half of the members are present.
 In case of extended absence, the organization of the absent member may appoint a
```
pmorie, line 109: I meant to remove this; will push an update.

STEERING-COMMITTEE.md
```diff
@@ -143,14 +109,6 @@ Quorum is considered reached when at least half of the members are present.
...
 single delegate from the same company during the absence.
```
mattmoor, line 110: Should this be updated?

mattmoor:
Generally, I feel as though the process we followed for TOC elections was well received by the community, and I really like the idea of following a similar model here.  I'd echo Doug's concerns around bias in candidacy towards folks making technical contributions, but I think that reflects a bias in the stats we have available today, and folks should feel comfortable leveraging the exception process (this is one of those things we'll necessarily improve over time).  To borrow a term widely used in kube, I think what's key is that candidates demonstrate a willingness to "chop wood and carry water" within the community, which feels like a staple of virtually all of our project leadership roles.

thanks for putting this together!

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+### Eligibility to Vote
+
+Anyone who has at least 50 contributions in the last 12 months is eligible to
+vote in the Steering election. Contributions are defined as opening PRs,
+reviewing and commenting on PRs, opening and commenting on issues, writing
+design docs, commenting on design docs, helping people on slack, participating
+in working groups, and other efforts that help advance Knative the project.
+
+[This
+dashboard]((https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year)
```
mattmoor, line 168:
```suggestion
dashboard](https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year)
```

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+## Transition
+
+In order to transition from the original, organization-appointed Steering
+Committee structure, elections will be phased in over 14 months.
+
+In October 2020, four positions will be elected, including three Contributing
+seats and one End-User seat.  One Google, one Red Hat/IBM, and one VMWare seat
+will be retained from the current Steering Committee.
+
+In October or November 2021, the three "old SC" seats will be replaced by
```
mattmoor, line 253:
```suggestion
In October or November 2021, the remaining three "bootstrap SC" seats will be replaced by
```

pmorie:
@duglin 

>  IOW, can we just merge the two groups into one?

I definitely think that's an option worth considering. Any ideas about how combining them could work?

jberkus:
@mattmoor FWIW, one of the things we did in Kubernetes was to track all events, management tasks, mentoring, etc. via GH repos.  This has made it a lot easier for us to capture non-code/docs contributions automatically via the stats.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+organization, sufficient members of the committee must resign until only two
+employees of the same employer are left. If it is impossible to find sufficient
+members to resign, all employees of that organization will be removed and new
+special elections held. In the event of a question of company membership (for
+example evaluating independence of corporate subsidiaries) a majority of all
+non-involved Steering Committee members will decide.
+
+## Vacancies
+
+In the event of a resignation or other loss of an elected SC member, the
```
n3wscott, line 232: What is the process for someone on steering who changes companies and remains on steering? That move might violate the company limit.

What about if a company is acquired and the new union now violate the limit?

pmorie: See line 214: https://github.com/knative/community/pull/214/files#diff-885f133673bb843be731122e9098f958R221

markusthoemmes:
In the realm I can grasp this in, this looks great to me! I'll trust others ask the hard-to-answer questions I can't think of. Thanks @pmorie

evankanderson:
> @pmorie thanks for this. Question: give the list of SC members are more technical in nature than today's SC members (not implying anything :-) other than today SC folks are not necessarily required to be contributors), I think this makes the line between the SC and TOC kind of blurry and it makes me wonder why we'd need two layers within a project of this size. IOW, can we just merge the two groups into one?

I think there may be a disposition difference between Steering and TOC in terms of contributions. Historically, Knative has had much stronger technical direction than "product direction". (I know that talking about as OSS project as a product bothers some people, but I think the CNCF has shown that this can be very effective in building a larger community.)

I see TOC as primarily focused on the technical quality of project; in some cases, this includes debugging organizational issues in the service of better technical outcomes, but the focus is on the technical outcome.

IMO, Steering should provide a complementary set of services around the non-technical aspects of the project (historically, the investment here has been lighter, but I think this proposal aims to change that somewhat). Some areas that I'd expect Steering to weigh in on:

* Meetups, polls, newsletters, etc. Possibly directly organizing these efforts, evolving into a framework if the effort gets too large.
* Tracking and managing relationships with large consumers of Knative, whether they be vendors or self-hosted installs.
* Collecting and organizing community feedback and helping determine which areas of functionality are most critical for the project to execute on. (TOC is concerned about executing well, but may not have the insight on _where_ to improve.)
* Tracking community health and coordinating conference presences, etc.
* Managing resource acquisition for all the tools that the project uses -- slack, CI/CD execution environments, website hosting, etc.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +120,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+
+Steering Committee members will serve for 2-year terms, staggered in order to
+preserve continuity.  Every year either two or three contributor seats will be
+elected, and one end user seat will be elected.
+
+## Election Procedure
+
+### Timeline
+
+Steering Committee elections shall be held in October or November of each year.
```
evankanderson, line 135: Do we want to suggest that this is intentionally about 6 months off from the TOC to allow better knowledge transfer and continuity?

STEERING-COMMITTEE.md
mattmoor: I'd expect same as TOC, someone would step down and backfill logic would kick in, is it not covered below?

STEERING-COMMITTEE.md
```diff
@@ -162,13 +120,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+
+Six weeks or more before the election, the Steering Committee will appoint
+between one and three Election Officer(s) to administer the election.  Elections
+Officers will be Knative community members in good standing who are eligible to
+vote, are not running for Steering in that election, and can make a public
+promise of impartiality.  They will be responsible for:
+
+* Making all announcements associated with the election
+* Preparing and distributing electronic ballots
+* Judging exception requests
```
evankanderson, line 155: Not sure if we want to send these to steering as a feedback loop, or keep it with the 1-3 election officers.

STEERING-COMMITTEE.md
mattmoor: ah, it's above!
> If employers change because of job changes, acquisitions, or other events, in a way that would yield more than 2 seats being held by employees of the same organization, sufficient members of the committee must resign until only two employees of the same employer are left.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +120,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+* Assisting candidates in preparing and sharing statements
+* Tallying voting results according to the rules in this charter
+
+### Eligibility to Vote
+
+Anyone who has at least 50 contributions in the last 12 months is eligible to
+vote in the Steering election. Contributions are defined as opening PRs,
+reviewing and commenting on PRs, opening and commenting on issues, writing
+design docs, commenting on design docs, helping people on slack, participating
+in working groups, and other efforts that help advance Knative the project.
```
evankanderson, line 165: Not sure how we'd want to count other community effort around things like relaying customer feedback to the right WGs, organizing community events, etc.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +120,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+
+We are committed to an inclusive process and will adapt future eligibility
+requirements based on community feedback.
+
+### Candidate Eligibility
+
+Community members must be eligible to vote in order to stand for election (this
+includes voters who qualify for an exception). Candidates may self-nominate or
+be nominated by another eligible member. The approximate time commitment of a
+Steering member is around 5 hours per week.
```
evankanderson, line 192: Is this true for steering? It looks copy-pasted from TOC, so want to verify.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +120,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+
+To run for a Contributing seat, a candidate must additionally be at least a
+project Member as defined in Roles.md.
+
+To run for an End-User seat, a candidate must be employed by, or own, an
+organization that does not sell products or services derived from Knative.
+Staff with in-house Knative deployments, as well as application developers who
+create applications that run on Knative, are eligible (and encouraged to run)
+for these seats.  Additionally, the candidate must participate in the
+knative-users mailing list and/or community-facing Knative Slack channels.
```
evankanderson, line 202: It's a little odd to have this in combination with the "must be able to vote" requirement.

I'm willing to give it a try, but it might make sense to specifically call out that candidates here should feel welcome to apply for an exception.

duglin:
> Any ideas about how combining them could work?

In my simple mind... just do it :-)  Not being on the SC I'm not sure all of the things it does, but it doesn't seem too onerous, and so just move those tasks into the TOC's job description.

re: Evan's take on the split between SC and TOC
If that's the split people have in mind, then I'm not sure it's an SC as much as an "process/mangement wg" type of thing.  But it's not something that sits on top of the TOC from an org perspective.

To be clear, I'm just looking for ways to flatten the org... I'm not a fan of extra "process" and "bureaucracy" and I see most work being done by the WGs (and quite well) so I'd prefer if most decisions were at the WG level, leaving the exceptions to the TOC.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +120,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+Elections will be held using a time-limited
+[Condorcet](https://en.wikipedia.org/wiki/Condorcet_method) ranking on
+[CIVS](http://civs.cs.cornell.edu/) using the Schulze method. The top
+vote-getters will be elected to the open seats, with the exceptions for company
+representation discussed below. This is the same process used for TOC elections.
+
+### Limitations on Company Representation
+
+No more than two seats may be held by employees of the same organization (or
+consortium, in the case of companies owning each other).  Additionally, the two
```
mswilson, line 215: I suggest "conglomerate" (as a grouping of companies with parent / subsidiary relationship) instead of "consortium" (an association of independent organizations).

STEERING-COMMITTEE.md
bsnchan: We approximated 5 hours, since steering specific tasks will be distributed across 7 members.

STEERING-COMMITTEE.md
pmorie: That's a good point, thanks! I'll make a change.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+
+We are committed to an inclusive process and will adapt future eligibility
+requirements based on community feedback.
+
+### Candidate Eligibility
+
+Community members must be eligible to vote in order to stand for election (this
+includes voters who qualify for an exception). Candidates may self-nominate or
+be nominated by another eligible member. The approximate time commitment of a
+Steering member is around 5 hours per week.
```
jchesterpivotal, line 192: I think this sentence could be left off. It's advice to potential candidates about the role, rather than a rule.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
 2. Open an issue on a knative repository and indicate that you would like
    attention from the steering committee.
 
+## Composition
+
+The steering committee will have 7 seats.  Five of the seats will be
```
jchesterpivotal, line 122: It'd be worth working out the use of tenses. "Will have" future tense or "has" present tense. I'd suggest present tense, it's always easier to read.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+Steering Committee members will serve for 2-year terms, staggered in order to
+preserve continuity.  Every year either two or three contributor seats will be
+elected, and one end user seat will be elected.
+
+## Election Procedure
+
+### Timeline
+
+Steering Committee elections shall be held in October or November of each year.
+Six weeks or more before the election, the Steering Committee shall appoint
```
jchesterpivotal, line 136: Continued from my note above: I'd recommend against "shall appoint", which is a future tense. I'd suggest "appoints".

If you keep future tenses, I'd use "will" over "shall". The latter is common in legalese, which can make things less approachable.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+
+Community members must be eligible to vote in order to stand for election (this
+includes voters who qualify for an exception). Candidates may self-nominate or
+be nominated by another eligible member. The approximate time commitment of a
+Steering member is around 5 hours per week.
+
+To run for a Contributing seat, a candidate must additionally be at least a
+project Member as defined in Roles.md.
+
+To run for an End-User seat, a candidate must be employed by, or own, an
```
jchesterpivotal, line 197: Does ownership include partial ownership via shares or partnership agreement?

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+
+In the event of a resignation or other loss of an elected SC member, the
+candidate with the next most votes from the previous election will be offered
+the seat, provided that person otherwise qualifies to join the SC. This process
+will continue until the seat is filled.
+
+In case this fails to fill the seat, a special election for that position will
+be held as soon as possible, unless the regular SC election is less than 7 weeks
+away.  Eligible voters from the most recent election will vote in the special
+election (ie: eligibility will not be redetermined at the time of the special
```
jchesterpivotal, line 240: I'd change the parenthetical to be inline: "Eligible voters from the most recent election will vote in the special election. Eligibility is not redetermined for the special election."

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+we value. We expect this metric not to capture everyone who should be eligible
+to vote. If a community member has had significant contributions over the past
+year but is not captured in the [stats.knative.dev
+dashboard](https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year),
+they will be able to submit an exception form to the Elections Officer(s) who
+will then review and determine whether this member should be marked as an
+exception.  All exceptions, and the reasons for them, will be recorded in a log
+that will be available to Steering and the TOC.
+
+All eligible voters will be captured at
```
jchesterpivotal, line 179: I'd suggest calling this an electoral roll.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+knative/community/steering-elections/$YEAR/voters.md and the voters’ guide will
+be captured at knative/community/steering-elections/$YEAR/README.md, similar to
+the kubernetes election process and identical to the TOC election process.
```
jchesterpivotal, line 182: I suggest "... Kubernetes election TOC process and identical to the Knative TOC election process".

(I assume the reference here is to the Kubernetes TOC).

pmorie: it's the kube steering election process, I see that it might be ambiguous now

STEERING-COMMITTEE.md
rhuss: Assuming that TOC's and steering's tasks are not overlapping, the time commitment might be also an argument to not merge TOC and steering (as discussed above) since this would increase the time commitment to ~ 11 hours in average per member (assuming 7 seats for a combined board, (8 * 5 + 5 * 7) / 7 = 10.7)

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+
+We are committed to an inclusive process and will adapt future eligibility
+requirements based on community feedback.
+
+### Candidate Eligibility
+
+Community members must be eligible to vote in order to stand for election (this
```
rhuss, line 189: Is it possible for an existing steering member to immediately run again for a seat? Or is there a cadence period of e.g. 1 year until they can be reelected? Couldn't find any information on this in the doc (but could have overlooked it, too)

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+for the person they are replacing, regardless of the length of that remainder.
+
+## Transition
+
+In order to transition from the original, organization-appointed Steering
+Committee structure, elections will be phased in over 14 months.
+
+In October 2020, four positions will be elected, including three Contributing
+seats and one End-User seat.  One Google, one Red Hat/IBM, and one VMWare seat
+will be retained from the current Steering Committee.
```
rhuss, line 251: Maybe add a brief comment, that for the transition period 2020/21 that there will be 6 contributing seats and 1 End-User seat for this transition phase.

STEERING-COMMITTEE.md
bsnchan: Definitely not a hard rule but we wanted to be clear about time commitment so folks 1. make sure they have the time bandwidth to dedicate to steering, and 2. other steering members can expect that sort of time commitment as well. This is to provide the guardrails so that responsibilities and tasks can be equally distributed between steering members.

STEERING-COMMITTEE.md
```diff
@@ -37,8 +38,6 @@ the way we run this committee, based on feedback from the community.
    disposition of those resources.
 1. Manage the Knative brand and decide which things can be called "Knative" and
    how that mark can be used in relation to other efforts or vendors.
-1. Confirm/reject nominations to the KSC from organizations who are allocated
-   seats.
 1. Confirm/reject nominations to the Technical Oversight Committee.
```
julz, line 41: Just to play devil's advocate a little here: why should the Steering Committee have the ability to reject a nomination to the TOC when the TOC is elected on essentially the same basis, from just as representative a pool, as the Steering committee? (in fact, almost by definition a TOC election being considered by the Steering Committee would be _more_ representative, in that it would obviously be more recent if it were being confirmed or rejected by the existing body).

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+
+All eligible voters will be captured at
+knative/community/steering-elections/$YEAR/voters.md and the voters’ guide will
+be captured at knative/community/steering-elections/$YEAR/README.md, similar to
+the kubernetes election process and identical to the TOC election process.
+
+We are committed to an inclusive process and will adapt future eligibility
+requirements based on community feedback.
+
+### Candidate Eligibility
```
julz, line 187: Im not sure if I missed it (sorry if I did :)), but I can't see here whether we've specified whether someone can be on both the TOC and the Steering committee concurrently? Can a TOC member run? If they can and they win are the expected to step down? etc

STEERING-COMMITTEE.md
mattmoor: I think better wording here now would be "Serve as TOC election officers" (linking to election mechanics), since I believe the TOC election mechanics say something similar.  

Happy to discuss that detail in more depth (I had the same thought), but I don't want to detract from the substance of the PR.  😇

STEERING-COMMITTEE.md
mattmoor: Good question!  I've wondered the same before and am mixed on this, but curious what prior art there is, since I could make a case for either. 😅

STEERING-COMMITTEE.md
mattmoor: If we opt for a cooldown period, I'd propose that we exempt bootstrap steering from the cooldown, since some will necessarily have to step down.  I'd be curious what other projects have done too.  My $0.02

STEERING-COMMITTEE.md
julz: Yep, if this becomes just an election officer responsibility rather than this confirm/reject wording that seems fine. 

I do think though fwiw that this actually cuts quite directly at a major hesitation I have about this charter (which generally I support) in a more general sense. How does having both TOC and Steering elected change the relationship between those bodies? For example this document defines the TOC as being a delegation of the Steering committee, presumably implying steering can override TOC decisions, or abolish the TOC.. is that still reasonable when both are elected bodies? Why? In what sense does, should or can Steering posses greater legitimacy?

mattmoor: re: combining

I think the groups ultimately cater to different candidate-audiences (perhaps overlapping) because they generalize exercise different skills operating in different aspects of the project.

I could see combining them dissuading candidates from joining the body, or elections for a single body resulting in a bias towards one skillset and a deficit of the other.  Don't ask me to walk and chew gum, @julz! 😉 

---

re: hierarchy

I think some of this wording is prehistoric and should probably be revisited.  Ideally steering (as election officers) would be responsible for tweaking election mechanics, and it's a good question whether more extreme capabilities are still needed.

Personally, I would mind this clause less if Steering were also elected from the community 😇

STEERING-COMMITTEE.md
julz: yah, I feel like it'd be pretty odd to have to think "hmm better not vote for Matt for TOC since then he can't be on Steering" ;). On the other hand, allowing people to be on both committees seems to make having two committees pretty weird (+ if Steering is supposed to be an escalation point for TOC, which it currently is in this doc, how does that even work when many members are the same?). 

I personally lean towards letting people who want to be on both be on both, but either way, I feel like we should be explicit about this, so very interested in opinions and precedents also!

STEERING-COMMITTEE.md
julz: To state the somewhat obvious, the problem with two committees _without_ hierarchy is that you have no obvious mechanism for how to resolve conflicts, and conflicts happen. The problem with two committees _with_ hierarchy when both are elected from essentially the same pool of candidates and electors it there's a very clear legitimacy problem when one overrules the other, particularly if the "junior" committee happens to've been more recently elected.

julz: (for the record I 100% believe in your ability to both walk and chew gum @mattmoor, but I of course respect any decision you make not to ;))

STEERING-COMMITTEE.md
sebgoa: agree, you can delete the last sentence.

STEERING-COMMITTEE.md
sebgoa: agree, this feels a bit hard to qualify for an end-user seat: you need to use, but not sell and contribute..

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+
+Community members must be eligible to vote in order to stand for election (this
+includes voters who qualify for an exception). Candidates may self-nominate or
+be nominated by another eligible member. The approximate time commitment of a
+Steering member is around 5 hours per week.
+
+To run for a Contributing seat, a candidate must additionally be at least a
+project Member as defined in Roles.md.
```
sebgoa, line 195: link to roles.md

STEERING-COMMITTEE.md
sebgoa: write VMWare/Pivotal if you write Red Hat/IBM

julz: does the Pivotal brand name still exist? I'd think it'd be "VMWare/Tanzu" if anything now?

STEERING-COMMITTEE.md
```diff
@@ -37,8 +38,6 @@ the way we run this committee, based on feedback from the community.
    disposition of those resources.
 1. Manage the Knative brand and decide which things can be called "Knative" and
```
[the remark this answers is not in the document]
jchesterpivotal: My hunch, though I am not a lawyer, is that we have created an unincorporated association with Google as the trustee of the Knative trademarks. That's generally an OK thing for the community, since we have these rules agreed to.

STEERING-COMMITTEE.md
vaikas: > Yep, if this becomes just an election officer responsibility rather than this confirm/reject wording that seems fine.
> 
> I do think though fwiw that this actually cuts quite directly at a major hesitation I have about this charter (which generally I support) in a more general sense. How does having both TOC and Steering elected change the relationship between those bodies? For example this document defines the TOC as being a delegation of the Steering committee, presumably implying steering can override TOC decisions, or abolish the TOC.. is that still reasonable when both are elected bodies? Why? In what sense does, should or can Steering posses greater legitimacy?

Seems like the last bullet point in this charter is what's being discussed here? The way I read this:
```
1. Act as the final escalation point and decider for any disputes, issues,
   clarifications, or escalations within the project scope.	   clarifications, or escalations within the project scope.
```

Is that if TOC makes a decision, it gets escalated to SC they would have the authority to override that decision?

STEERING-COMMITTEE.md
```diff
@@ -60,28 +59,26 @@ committee currently recognizes this delegated authority for:
...
 
-Given the private nature of many of these discussions (e.g. privacy, private
+Given the private nature of some of these discussions (e.g. privacy, private
 emails to the committee, code of conduct violations, escalations, disputes
-between members, security reports, etc.) meetings are held in private.
+between members, security reports, etc.) some meetings are held in private.
 
-Meeting notes are available to members of the knative-dev mailing list (link to
-be added).
+Meeting notes are available to members of the [knative-dev mailing list](https://groups.google.com/forum/#!forum/knative-dev).
```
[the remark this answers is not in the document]
bsnchan: Yep! Recording will be made available on the Knative YouTube channel!

STEERING-COMMITTEE.md
pmorie: Incorporated in next push

STEERING-COMMITTEE.md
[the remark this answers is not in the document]
pmorie: I recognize that this isn't currently followed, consider following it part of the proposal.

STEERING-COMMITTEE.md
pmorie: The material on the TOC elections currently doesn't say what time of year it will be held, I wonder if we should just say "annually" ? WDYT?

STEERING-COMMITTEE.md
pmorie: I don't think exception requests will normally be difficult to judge, and I would trust election officers to do so personally. Do you think any text should change here?

STEERING-COMMITTEE.md
mattmoor: I'd leave it and update TOC.  It'd be good to indicate when that happens for folks looking to understand how they might participate.

STEERING-COMMITTEE.md
pmorie: I feel like these are covered by the presence of the exception process, do you think they need to be explicitly called out?

STEERING-COMMITTEE.md
pmorie: I removed the sentence in question :)

STEERING-COMMITTEE.md
pmorie: Agree that it's a little odd. Thinking about how to phrase this.

STEERING-COMMITTEE.md
pmorie: Agree, and I'm changing it wherever it makes sense (the language about elections seems appropriate to use future tense in since it is describing the next election, which hasn't happened yet at any given time)

STEERING-COMMITTEE.md
pmorie: Ack, incorporating now

STEERING-COMMITTEE.md
pmorie: Took the 'own' language out because I felt it was clearer that way.

STEERING-COMMITTEE.md
pmorie: ack

STEERING-COMMITTEE.md
pmorie: I added language explicitly saying there is no cooldown period. I would prefer to discuss term limits / cooldown as future changes to the charter if this proposal is adopted.

STEERING-COMMITTEE.md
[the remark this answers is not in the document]
pmorie: You raise a valid question, I'm not sure I have the brain to answer it. I'll think it over, just want to explicitly ack it.

STEERING-COMMITTEE.md
```diff
@@ -60,28 +59,26 @@ committee currently recognizes this delegated authority for:
...
 
-The committee was created as the project was in its infancy, in order to tackle
-governance and overall project strategy. Because of the nature of the project
-and funding required, it was decided that strong corporate leadership was
-necessary for the project to ensure velocity. As the project grows and matures
-the KSC will, from time to time, consider if this policy should be changed.
+When the Knative project was created, it was assigned strong corporate
+leadership through organizationally appointed steering committee members. As a
+mature project with production adoption, the project has moved to a democratic
+process of selecting new SC members.  This process may change again in the
```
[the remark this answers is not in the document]
pmorie: ack

STEERING-COMMITTEE.md
```diff
@@ -60,28 +59,26 @@ committee currently recognizes this delegated authority for:
...
+future as Knative continues to grow.
```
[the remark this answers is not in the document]
pmorie: ack

STEERING-COMMITTEE.md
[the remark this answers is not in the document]
pmorie: Yes, public meetings will be recorded and the recordings made available, I'll add this language

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
...
+In October 2020, four positions will be elected, including three Contributing
+seats and one End-User seat.  One Google, one Red Hat/IBM, and one VMWare seat
+will be retained from the current Steering Committee.
+
+In October or November 2021, the three "old SC" seats will be replaced by
+elected seats (two Contributing and one End-User).
+
+## Changes to the charter
+
+Changes to the KSC charter may be proposed via a Pull Request on the charter
```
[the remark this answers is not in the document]
pmorie: ack

pmorie:
I pushed new changes incorporating some feedback, still thinking about some of the good but challenging questions raised here. Thanks for all the participation so far!

STEERING-COMMITTEE.md
```diff
@@ -119,28 +120,28 @@ There are two ways to raise issues to the steering committee for decision:
 
 ## Composition
 
-The steering committee will have 7 seats.  Five of the seats will be
-Contributing seats, open to any project contributor.  Two of the seats will be
-End-User seats, open only to active end-user community members.  See Candidate
+The steering committee has 7 seats.  Five of the seats will be Contributing
+seats, open to any project contributor.  Two of the seats will be End-User
```
jchesterpivotal, line 124: Mismatched tenses: "has" vs "will be".

STEERING-COMMITTEE.md
evankanderson: The feedback loop was to make steering aware of the sorts of work that we weren't currently judging well, but I'm fine either way.

STEERING-COMMITTEE.md
evankanderson: To agree with julz's point, above, there needs to be a clear ultimate authority. I would prefer that authority rest with the community part of the committees, rather than the technology part of the committees. In particular, the Steering Committee is the group which adjudicates Code of Conduct complaints, and I'd prefer that those decisions to override *any* technical considerations.

STEERING-COMMITTEE.md
evankanderson: I'm also assuming that a candidate could step down at any moment, or conditionally resign from one committee in order to serve on the other if elected.

I do think serving on both committees might be a bit much, but I'm not sure I'd want to explicitly prohibit it.

STEERING-COMMITTEE.md
```diff
@@ -60,28 +59,27 @@ committee currently recognizes this delegated authority for:
 
 KSC meets every two weeks, or as-needed. Meetings are held online.
 
-Given the private nature of many of these discussions (e.g. privacy, private
+Given the private nature of some of these discussions (e.g. privacy, private
 emails to the committee, code of conduct violations, escalations, disputes
-between members, security reports, etc.) meetings are held in private.
+between members, security reports, etc.) some meetings are held in private.
```
csantanapr, line 64: I would like to see a policy that states that by default all meetings will be publicly available, that exceptions would be only when meeting to discuss privacy/private matters. Is ok to have held meetings and split the time between public and private portion, the private portion will not be recorded.

STEERING-COMMITTEE.md
```diff
@@ -60,28 +59,27 @@ committee currently recognizes this delegated authority for:
...
 
-Meeting notes are available to members of the knative-dev mailing list (link to
-be added).
+Meeting notes are available to members of the [knative-dev mailing
+list](https://groups.google.com/forum/#!forum/knative-dev). Public meetings will
+be recorded and the recordings made available publicly.
```
csantanapr, line 69: We have two mailing list knative-dev and knative-users, wondering if the steering community should have more visibility to the knative-users mailing list

STEERING-COMMITTEE.md
```diff
@@ -60,28 +59,27 @@ committee currently recognizes this delegated authority for:
...
 [issues in the docs repo](https://github.com/knative/docs/issues), and the KSC
-invites your feedback there. See [Getting in touch](#getting-in-touch) for other
+invites your feedback there. See [Raising Issues](#raising-issue-to-the-steering-committee) for other
 options.
 
 ## Committee members
 
-Seats on the Steering Committee are held by an organization, not by the
-individual.
+Seats on the Steering Committee are held by an individual, not by their employer.
```
csantanapr, line 77: This is shortest line on this PR and the strongest statement of open source values. 💯

STEERING-COMMITTEE.md
```diff
@@ -60,28 +59,27 @@ committee currently recognizes this delegated authority for:
...
 
-The committee was created as the project was in its infancy, in order to tackle
-governance and overall project strategy. Because of the nature of the project
-and funding required, it was decided that strong corporate leadership was
-necessary for the project to ensure velocity. As the project grows and matures
-the KSC will, from time to time, consider if this policy should be changed.
+When the Knative project was created, it was assigned strong corporate
```
csantanapr, line 79: >When the Knative project was created, it was assigned strong corporate
leadership through organizationally appointed steering committee members.

I would leave out this sentence don't think it adds any "community" value this type of historical info.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +118,158 @@ There are two ways to raise issues to the steering committee for decision:
...
+dashboard](https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year),
+they will be able to submit an exception form to the Elections Officer(s) who
+will then review and determine whether this member should be marked as an
+exception.  All exceptions, and the reasons for them, will be recorded in a log
+that will be available to Steering and the TOC.
+
+The electoral roll of all eligible voters will be captured at
+knative/community/steering-elections/$YEAR/voters.md and the voters’ guide will
+be captured at knative/community/steering-elections/$YEAR/README.md, similar to
+the kubernetes election process and identical to the TOC election process.
```
csantanapr, line 183: >kubernetes election process

Can make this a link?

STEERING-COMMITTEE.md
```diff
@@ -162,13 +118,158 @@ There are two ways to raise issues to the steering committee for decision:
...
+in working groups, and other efforts that help advance Knative the project.
+
+[This
+dashboard]((https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year)
+shows only GitHub based contributions and does not capture all the contributions
+we value. We expect this metric not to capture everyone who should be eligible
+to vote. If a community member has had significant contributions over the past
+year but is not captured in the [stats.knative.dev
+dashboard](https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year),
+they will be able to submit an exception form to the Elections Officer(s) who
```
csantanapr, line 175: the individual or a WG lead. I think WG leads would notice that an individual that is contributing that deserves to be included for voting, so the WG could be the one submitting the form.

STEERING-COMMITTEE.md
```diff
@@ -162,13 +118,158 @@ There are two ways to raise issues to the steering committee for decision:
...
+over 14 months.
+
+In October 2020, four positions will be elected, including three Contributing
+seats and one End-User seat. One Google, one IBM, and one VMWare seat will be
+retained from the current Steering Committee.
+
+In October or November 2021, the three remaining "bootstrap SC" seats will be replaced by
+elected seats (two Contributing and one End-User).
+
+## Changes to the charter
```
whaught, line 258:
```suggestion
## Changes to the Charter
```

STEERING-COMMITTEE.md
```diff
@@ -162,13 +118,158 @@ There are two ways to raise issues to the steering committee for decision:
...
+two weeks and not more than four.  The results of the election will be
+announced within one week of closing the election.  New Steering Committee
+members will take office on December 1 of each year.
+
+### Election Officer(s)
+
+Six weeks or more before the election, the Steering Committee will appoint
+between one and three Election Officer(s) to administer the election.  Elections
+Officers will be Knative community members in good standing who are eligible to
+vote, are not running for Steering in that election, and can make a public
```
lance, line 151: It's not clear to me from this paragraph if standing Steering Committee members may also be Election Officers

STEERING-COMMITTEE.md
```diff
@@ -162,13 +118,158 @@ There are two ways to raise issues to the steering committee for decision:
...
+[This
+dashboard]((https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year)
+shows only GitHub based contributions and does not capture all the contributions
+we value. We expect this metric not to capture everyone who should be eligible
+to vote. If a community member has had significant contributions over the past
+year but is not captured in the [stats.knative.dev
+dashboard](https://knative.teststats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year),
+they will be able to submit an exception form to the Elections Officer(s) who
+will then review and determine whether this member should be marked as an
+exception.  All exceptions, and the reasons for them, will be recorded in a log
```
lance, line 177:
```suggestion
will then review and determine whether this member should be eligible to vote.
All exceptions, and the reasons for them, will be recorded in a log
```

STEERING-COMMITTEE.md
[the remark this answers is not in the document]
duglin: While we're in here, can we add "Logo" into this bullet point since they control that too?

STEERING-COMMITTEE.md
```diff
@@ -60,28 +59,27 @@ committee currently recognizes this delegated authority for:
...
+between members, security reports, etc.) some meetings are held in private.
 
-Meeting notes are available to members of the knative-dev mailing list (link to
-be added).
+Meeting notes are available to members of the [knative-dev mailing
+list](https://groups.google.com/forum/#!forum/knative-dev). Public meetings will
+be recorded and the recordings made available publicly.
 
 Questions and proposals for changes to governance are posted as
 [issues in the docs repo](https://github.com/knative/docs/issues), and the KSC
```
duglin, line 71: Should this be "community" repo since this doc is in that repo, not "docs" ?

STEERING-COMMITTEE.md
```diff
@@ -140,17 +107,6 @@ pass with majority of the committee supporting it.
```
duglin, line 107: to be clear, it's: `> 50% of votes cast`, not of all members, right? And, abstain == no vote

jberkus: It's actually 51% of all members, the way the SC charter is currently written. Paul's PR doesn't change that.

duglin: Yes but I think it should be changed, otherwise absent, or abstaining, members influence the vote

STEERING-COMMITTEE.md
evankanderson: I suspect that Paul means that the notes will be in the team drive, which would grant read access to knative-users as well.

STEERING-COMMITTEE.md
duglin: I'm ok with the "at least half of the members" in the next sentence.

STEERING-COMMITTEE.md
pmorie: Just a note: we decided to have a distinct discussion about the relationship of the SC/TOC, dev list thread here:

https://groups.google.com/g/knative-dev/c/AOm2exWqjJg

STEERING-COMMITTEE.md
pmorie: That's correct and what I had meant. I'll make it explicit.

STEERING-COMMITTEE.md
pmorie: I don't disagree but I am going to refrain from incorporating that for now in order to keep the focus on the most important parts. Hope it's okay to revisit that later on.

STEERING-COMMITTEE.md
pmorie: Good point

STEERING-COMMITTEE.md
pmorie: Fair, I think that's context that lives in this PR just fine.

STEERING-COMMITTEE.md
pmorie: ack, thanks!

STEERING-COMMITTEE.md
```diff
@@ -162,13 +117,157 @@ There are two ways to raise issues to the steering committee for decision:
 2. Open an issue on a knative repository and indicate that you would like
    attention from the steering committee.
 
+## Composition
+
+The steering committee will have 7 seats.  Five of the seats will be
+Contributing seats, open to any project contributor.  Two of the seats will be
+End-User seats, open only to active end-user community members.  See Candidate
+Eligibility for definitions of these categories.
```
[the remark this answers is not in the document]
pmorie: ack, thanks!

STEERING-COMMITTEE.md
pmorie: Good point, I will incorporate this explicitly

STEERING-COMMITTEE.md
pmorie: My preference would be for people to explain their own exception requests, unless there's a strong group-level preference for proxy exception requests

STEERING-COMMITTEE.md
pmorie: Ack

STEERING-COMMITTEE.md
pmorie: I don't think it should be prohibited. In the event someone is qualified, wants to do both roles, and the community wants it, why not?

STEERING-COMMITTEE.md
pmorie: Ack, thanks

STEERING-COMMITTEE.md
```diff
@@ -162,13 +114,161 @@ There are two ways to raise issues to the steering committee for decision:
...
 Portions of this document are adapted from the
 [Istio Steering Committee](https://github.com/istio/community/blob/master/STEERING-COMMITTEE.md)
-documentation, which is licensed under the Apache License 2.0.
+documentation, and other portions are copied from the [Kubernetes Steering Committee](https://github.com/kubernetes/steering/blob/master/elections.md), both of which are licensed under the Apache License 2.0.
 
 Except as otherwise noted, the content of this page is licensed under the
 [Creative Commons Attribution 4.0 License](https://creativecommons.org/licenses/by/4.0/),
 and code samples are licensed under the
-[Apache 2.0 License](https://www.apache.org/licenses/LICENSE-2.0).
+[Apache 2.0 License](https://www.apache.org/licenses/LICENSE-2.0).
```
mattmoor, line 274: Add trailing newline:
```suggestion
[Apache 2.0 License](https://www.apache.org/licenses/LICENSE-2.0).

```

STEERING-COMMITTEE.md
duglin: ^^ was w.r.t. quorum, not voting. Just to be clear.

re: voting....
- to me it need to be > 50% because if there's a tie then I think the status-quo should prevail
- we need to be clear about whether abstains and "not voting" counts towards the tally, otherwise people who choose to not vote (by not showing up, or voting to 'abstain') are in essence voting "no". Which means1) means we're allowing people who choose to step back from the process (and their duties) to influence things, and 2) there's no way for someone to say "I have no opinion and I'll go with the will of the other voters". That's why "abstain" should be the same as "not voting" and neither should count in the tally.  ie. "> 50% of people who vote"

STEERING-COMMITTEE.md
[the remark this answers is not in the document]
tiagocarmosantos: Hi, everyone, I'm new in this group and after reading this topic. I feel free to contribute with SC, cause I believe in the Knative project and in it future. So, i agree with @evankanderson opinion. I think that TOC and SC are different in disposition terms. One hand, TOC as a technical aspects, and Other hand a SC as non-technical aspects. When I say non-techincal aspects, is like the SC need focus on activies:

- Meetups, polls, newsletters, etc. Possibly directly organizing these efforts, evolving into a framework if the effort gets too large.

- Tracking and managing relationships with large consumers of Knative, whether they be vendors or self-hosted installs.

- Collecting and organizing community feedback and helping determine which areas of functionality are most critical for the project to execute on. (TOC is concerned about executing well, but may not have the insight on where to improve.)

- Tracking community health and coordinating conference presences, etc.

- Managing resource acquisition for all the tools that the project uses -- slack, CI/CD execution environments, website hosting, etc.

Like, Evans said on his comment. 
Finally, I guess this a good way to Mantained SC and support Knative community.

And to Formed this SC, I think it must lookfeel a "Consortium" like W3C to Steering the next steps in the next years.

matzew:
A good proposal!  Thanks for driving this!   LGTM

STEERING-COMMITTEE.md
jberkus: so, the way it is now (and is unchanged in this charter revision), a majority of SC needs to vote for any decision, and abstains are materially indisinguishable from "no" votes.  The advantage of the present system is mostly around preventing any one faction (or company, or whatever) from taking advantage of poor availability (say, during Kubecon) to pass changes without an actual majority.

I've been a part of organizations that allowed a minority to pass measures, but those organizations had very strict rules around what constituted a valid meeting in order to make that work -- rules I'll note that this charter lacks. That would be a lot of extra process for the SC.  What problem are you actually solving here?

duglin: - I want a way for someone to vote with "no opinion" - can't do that today
- I don't want people who choose to not participate to have influence over votes
- and yes I agree we probably need a bit more in here about quorum/valid meetings.  E.g. announce votes at least week in advance, allow people to vote off-line if they can't make the meeting, etc...  but in small groups like this it's usually not an issue when there's trust and good will.  However, yes we should add those things....

Create a charter assuming the worst, but hope for the best.

jberkus: Can I suggest passing this "simple" version first, and then considering that kind of revision?  The advantage of always requiring a majority affirmative vote is that it considerably simplifies process; you can take a vote anywhere you want because you always need four "yeses".

STEERING-COMMITTEE.md
pmorie: I decided the language about approving/rejecting nominations doesn't make a lot of sense, I am going to strike it.

STEERING-COMMITTEE.md
pmorie: I'm going to resolve this for now; we can come back to it once the key parts are resolved.

STEERING-COMMITTEE.md
duglin: that's fine, we can defer a deeper chat for later

STEERING-COMMITTEE.md
pmorie: I'm making it explicit that it's okay. Thanks for raising this @julz

STEERING-COMMITTEE.md
jberkus: FWIW, Kubernetes does it with the decisions made by the EC, with a record kept of each decision.

STEERING-COMMITTEE.md
geekygirldawn: Pivotal no longer exists as a standalone company. Previous Pivotal employees and the people working on Tanzu are all employees of VMware, so I think it's correct to just say "VMware" seat. Although, we should correct the capitalization of VMware :)

pmorie:
I have closed this as the conversation is currently centered on #277

bsnchan:
We've now merged in #277 which is based off of this proposal.

PR #246, opened 2020-08-28
add async repo to peribolos knative-sandbox.yaml

beemarie:


peribolos/knative-sandbox.yaml
```diff
@@ -468,6 +474,7 @@ orgs:
           net-istio: write
           net-kourier: write
           sample-controller: write
+          async-component: write
```
mattmoor, line 477: Please add it to the list for `admin` as well.

beemarie: @mattmoor done!

PR #249, opened 2020-08-31
Make it so the folks doing github actions exploring in sandbox can make releases

n3wscott:
Sandbox is intended to be a little less structured that core Knative. In working with github actions, I need to be able to make releases, but I am not a member of productivity. Making a new group to have admin over the `.github` and `github actions` related repos in `knative-sandbox`.

### DIFF
--- a/peribolos/knative-sandbox.yaml
+++ b/peribolos/knative-sandbox.yaml
@@ -480,13 +480,9 @@ orgs:
         repos:
           .github: admin
           downstream-test-go: admin
-        teams:
-          Git Hub Action Leads:
-            description: The leads for GitHub Actions for Knative
-            members:
-            - n3wscott
-            - mattmoor
-            privacy: closed
+        members:
+        - n3wscott
+        - mattmoor
       Serving Writers:
         description: Grants write access to serving-related repositories.
         privacy: closed
--- a/peribolos/knative-sandbox.yaml
+++ b/peribolos/knative-sandbox.yaml
@@ -474,6 +474,19 @@ orgs:
             - chizhg
             - chaodaiG
             privacy: closed
+      GitHub Actions Effort:
+        description: Grants admin rights to github actions related repositories.
+        privacy: closed
+        repos:
+          .github: admin
+          downstream-test-go: admin
+        teams:
+          Git Hub Action Leads:
+            description: The leads for GitHub Actions for Knative
+            members:
+            - n3wscott
+            - mattmoor
+            privacy: closed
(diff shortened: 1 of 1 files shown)


peribolos/knative-sandbox.yaml
```diff
@@ -474,6 +474,19 @@ orgs:
             - chizhg
             - chaodaiG
             privacy: closed
+      GitHub Actions Effort:
+        description: Grants admin rights to github actions related repositories.
```
evankanderson, line 478: Do you want to include a closing date on this?

peribolos/knative-sandbox.yaml
```diff
@@ -474,6 +474,19 @@ orgs:
...
+        privacy: closed
+        repos:
+          .github: admin
+          downstream-test-go: admin
+        teams:
```
evankanderson, line 483: I don't think you need a sub-team here, just include `members` in the upper "effort" directly.

peribolos/knative-sandbox.yaml
n3wscott: no end in sight

peribolos/knative-sandbox.yaml
n3wscott: ok I think I did, can you double check that?

peribolos/knative-sandbox.yaml
```diff
@@ -474,6 +474,15 @@ orgs:
             - chizhg
             - chaodaiG
             privacy: closed
+      GitHub Actions Effort:
+        description: Grants admin rights to github actions related repositories.
+        privacy: closed
+        repos:
+          .github: admin
+          downstream-test-go: admin
+        members:
```
n3wscott, line 483: is this correct @evankanderson ?

peribolos/knative-sandbox.yaml
evankanderson: When you and I and Matt have all gone on to other things, I'd like the next person cleaning up these ACLs to know what this was for. Any suggestions on the best way to record that?

(If it's time-bounded, they can just clean up after date X.)

peribolos/knative-sandbox.yaml
evankanderson: Yes.

peribolos/knative-sandbox.yaml
n3wscott: I added a comment above to help clarify why I am making this and why it is hard to just put a date in.

PR #275, opened 2020-09-14
Rejected: Steering Committee composition.

anicolao:
In response to issue #33.

Paul -- you had some ideas on how to incorporate elected seats. Feel free to add them as comments to the "composition" section so that we can iterate on that.

### DIFF
--- a/STEERING-COMMITTEE.md
+++ b/STEERING-COMMITTEE.md
@@ -7,51 +7,55 @@ aliases:
   - /contributing/steering-committee/
 ---
 
-The Knative Steering Committee (KSC) is the ultimate authority for the Knative
-project, and governs all aspects of the project.
+The Knative Steering Committee (KSC) defines the governance for Knative
+in pursuit of the goal of an open, vibrant project with a productive
+community, a healthy ecosystem of vendors, and most of all satisfied
+customers running their workloads on Knative clusters.
 
 The governance of Knative is an open, living document, and will continue to
 evolve as the community and project change. We expect over time we will adapt
 the way we run this committee, based on feedback from the community.
 
+- [Mission](#mission)
 - [Charter](#charter)
 - [Delegated authority](#delegated-authority)
 - [Committee Meetings](#committee-meetings)
 - [Committee Mechanics](#committee-mechanics)
 - [Committee Members](#committee-members)
-  - [Allocation of seats](#allocation-of-seats)
+  - [Composition](#composition)
+  - [Nomination of representatives](#nomination-of-representatives)
 - [Decision process](#decision-process)
 - [Changes to the charter](#changes-to-the-charter)
 - [Getting in touch](#getting-in-touch)
 
+## Mission
+
+The Steering Committee's mission is to guide the project such that:
+
+1. Knative is a truly open effort which is inclusive and empowers contributors
+1. Knative core is clearly defined with a conformance suite
+1. Trademark usage is monitored and violations addressed
+
 ## Charter
 
-1. Define, evolve, and promote the vision, values, mission, and scope of the
-   project.
-1. Define and evolve project governance structures and policies, including
-   project roles and how collaborators become members, approvers, leads, and/or
-   administrators. This includes policy for the creation and administration of
-   [working groups](./working-groups/WORKING-GROUPS.md) and committees.
-1. Steward, control access, delegate access, and establishes processes
-   regarding, all Knative project resources and has the final say in the
-   disposition of those resources.
-1. Manage the Knative brand and decide which things can be called "Knative" and
-   how that mark can be used in relation to other efforts or vendors.
+1. Define, evolve, and promote the values of the project, and address any
+   [Code of Conduct](./CODE-OF-CONDUCT.md) issues that may arise, and
+   maintain confidentiality.
+1. Review and approve changes to the scope of Knative core.
+1. Administer the marks in a transparent way, and take responsibility for
+   ensuring follow up on trademark violations.
 1. Confirm/reject nominations to the KSC from organizations who are allocated
    seats.
-1. Confirm/reject nominations to the Technical Oversight Committee.
-1. Receive and handle reports about [code of conduct](./CODE-OF-CONDUCT.md)
-   violations and maintain confidentiality.
-1. Receive security reports; work with the appropriate technical leads to accept
-   or reject the report; maintain the private nature of such reports until
-   disclosed to the broader community.
 1. Act as the final escalation point and decider for any disputes, issues,
    clarifications, or escalations within the project scope.
+1. Delegate all other responsibilities, including matters of governance
+   outside the scope of the items laid out above, to competent subcommittees
+   who are empowered to drive innovation and evolve the overall project.
 
 ## Delegated authority
 
-KSC may choose to delegate its authority to other committees as-needed. The
-committee currently recognizes this delegated authority for:
+KSC chooses to delegate as much of its authority as practical to community-run
+subcommittees. The committee currently recognizes this delegated authority for:
 
 - Technical guidance is delegated to the
   [Technical Oversight Committee](./TECH-OVERSIGHT-COMMITTEE.md).
@@ -60,9 +64,8 @@ committee currently recognizes this delegated authority for:
 
 KSC meets every two weeks, or as-needed. Meetings are held online.
 
-Given the private nature of many of these discussions (e.g. privacy, private
-emails to the committee, code of conduct violations, escalations, disputes
-between members, security reports, etc.) meetings are held in private.
+Given the private nature of most of these discussions (e.g. code of conduct violations, 
+escalations, legal concerns, etc.) meetings are held in private.
 
 Meeting notes are available to members of the knative-dev mailing list (link to
 be added).
@@ -76,12 +79,6 @@ options.
 
 Seats on the Steering Committee are held by an organization, not by the
 individual.
-
-The committee was created as the project was in its infancy, in order to tackle
-governance and overall project strategy. Because of the nature of the project
-and funding required, it was decided that strong corporate leadership was
-necessary for the project to ensure velocity. As the project grows and matures
-the KSC will, from time to time, consider if this policy should be changed.
 
 The current membership of the committee is currently (listed alphabetically by
 first name):
@@ -100,19 +97,25 @@ first name):
 \* Open seats have been allocated to the organization per the committee rules,
   but have not been assigned to a member of the organization.
 
-### Allocation of seats
+### Composition
 
-Seats on the steering committee are allocated based upon contribution to the
-project by an organization. No final decision has been made on the exact
-formula.
+As of 2020, the three organizations with the largest demonstrated 
+commitment to the project are Google, IBM/RedHat, and VMWare. For
+2021, a 5 person steering committee should be competent to
+execute its responsibilities, with three seats for Google, one seat
+for IBM/RedHat, and one seat for VMWare.
 
-As the project continues to grow, we expect to add additional seats to the
-committee, ensuring that those contributing to the project are properly
-represented.
+### Nomination of representatives
+
+Seats on the steering committee are allocated based upon lifetime
+contributions to the project by an organization. The intention is
+to ensure that those organizations are committed to the ongoing
+maintenance burden that adding to core implies, and express their
+commitment to the project and the project's users.
 
 - After a seat is allocated to an organization, the organization shall nominate
   a candidate to be confirmed by KSC. The committee reserves the right to not
-  confirm a candidate, in which the organization would need to nominate a new
+  confirm a candidate, in which case the organization would need to nominate a new
   candidate
 - Members of the committee may step down at any time. When a member steps down,
   their organization shall nominate a new candidate.
@@ -133,12 +136,15 @@ represented.
 
 ## Decision process
 
-The steering committee desires to always reach consensus.
+The steering committee desires to always reach consensus. The expectation
+is that lack of consensus requires multiple discussions and possibly public
+review, and that a vote would be an option of last resort.
 
-Decisions are made in meetings when a quorum of the members are present and may
-pass with majority of the committee supporting it.
+Decisions are made in meetings when a quorum of the members are present 
+and consensus is achieved, or in a meeting where all members are present
+and a vote is held and passed by majority.
 
-Quorum is considered reached when at least half of the members are present.
+Quorum is considered reached when a majority of the members are present.
 
 In case of extended absence, the organization of the absent member may appoint a
 single delegate from the same company during the absence.


STEERING-COMMITTEE.md
```diff
@@ -60,9 +64,8 @@ committee currently recognizes this delegated authority for:
 
 KSC meets every two weeks, or as-needed. Meetings are held online.
 
-Given the private nature of many of these discussions (e.g. privacy, private
-emails to the committee, code of conduct violations, escalations, disputes
-between members, security reports, etc.) meetings are held in private.
+Given the private nature of most of these discussions (e.g. code of conduct violations,
```
mattmoor, line 67: Remove trailing whitespace:
```suggestion
Given the private nature of most of these discussions (e.g. code of conduct violations,
```

STEERING-COMMITTEE.md
```diff
@@ -100,19 +97,25 @@ first name):
 \* Open seats have been allocated to the organization per the committee rules,
   but have not been assigned to a member of the organization.
 
-### Allocation of seats
+### Composition
+
+As of 2020, the three organizations with the largest demonstrated
```
mattmoor, line 102: Remove trailing whitespace:
```suggestion
As of 2020, the three organizations with the largest demonstrated
```

STEERING-COMMITTEE.md
```diff
@@ -133,12 +136,15 @@ represented.
...
 ## Decision process
 
-The steering committee desires to always reach consensus.
+The steering committee desires to always reach consensus. The expectation
+is that lack of consensus requires multiple discussions and possibly public
+review, and that a vote would be an option of last resort.
 
-Decisions are made in meetings when a quorum of the members are present and may
-pass with majority of the committee supporting it.
+Decisions are made in meetings when a quorum of the members are present
```
mattmoor, line 143: Remove trailing whitespace:
```suggestion
Decisions are made in meetings when a quorum of the members are present
```

STEERING-COMMITTEE.md
```diff
@@ -100,19 +97,25 @@ first name):
...
-formula.
+### Nomination of representatives
 
-As the project continues to grow, we expect to add additional seats to the
-committee, ensuring that those contributing to the project are properly
-represented.
+Seats on the steering committee are allocated based upon lifetime
+contributions to the project by an organization. The intention is
+to ensure that those organizations are committed to the ongoing
+maintenance burden that adding to core implies, and express their
```
julz, line 113: Does the intention to ensure organisations are committed to ongoing maintenance not seem a little at odds with a lifetime contribution criteria? If an organisation were no longer maintaining the project but did in the past, that seems like it'd be the opposite of demonstrating a commitment to ongoing maintenance. Can you see any argument for including at least an element of recent contribution, too?

STEERING-COMMITTEE.md
```diff
@@ -100,19 +97,25 @@ first name):
 \* Open seats have been allocated to the organization per the committee rules,
   but have not been assigned to a member of the organization.
 
-### Allocation of seats
+### Composition
+
+As of 2020, the three organizations with the largest demonstrated 
+commitment to the project are Google, IBM/RedHat, and VMWare. For
+2021, a 5 person steering committee should be competent to
+execute its responsibilities, with three seats for Google, one seat
```
julz, line 105: can you say a little more about the thinking behind the new committee going from 4:3 google/others to 3:2 google/others? Would it not seem a little unfortunate if a discussion about making steering more representative of the community ended up with it becoming _more_ dominated by one vendor?

STEERING-COMMITTEE.md
```diff
@@ -7,51 +7,55 @@ aliases:
...
 - [Changes to the charter](#changes-to-the-charter)
 - [Getting in touch](#getting-in-touch)
 
+## Mission
+
+The Steering Committee's mission is to guide the project such that:
+
+1. Knative is a truly open effort which is inclusive and empowers contributors
+1. Knative core is clearly defined with a conformance suite
+1. Trademark usage is monitored and violations addressed
```
evankanderson, line 37: I wonder if a more positive statement here (such as "participants have clear rules for usage of the Knative trademark") would make more sense as a mission than focusing on violatons.

evankanderson: As an example, everyone could agree that "this is the project that must not be named", which would meet this goal of the mission, but would clearly be a failure for the project.

STEERING-COMMITTEE.md
```diff
@@ -7,51 +7,55 @@ aliases:
...
    seats.
-1. Confirm/reject nominations to the Technical Oversight Committee.
-1. Receive and handle reports about [code of conduct](./CODE-OF-CONDUCT.md)
-   violations and maintain confidentiality.
-1. Receive security reports; work with the appropriate technical leads to accept
-   or reject the report; maintain the private nature of such reports until
-   disclosed to the broader community.
 1. Act as the final escalation point and decider for any disputes, issues,
    clarifications, or escalations within the project scope.
+1. Delegate all other responsibilities, including matters of governance
```
evankanderson, line 51: It's not clear that the charter needs to indicate whether the Steering Committee will delegate or handle other items directly; this seems like an operating policy for the committee, rather than something to put in the charter.

STEERING-COMMITTEE.md
```diff
@@ -7,51 +7,55 @@ aliases:
...
+   outside the scope of the items laid out above, to competent subcommittees
+   who are empowered to drive innovation and evolve the overall project.
 
 ## Delegated authority
 
-KSC may choose to delegate its authority to other committees as-needed. The
-committee currently recognizes this delegated authority for:
+KSC chooses to delegate as much of its authority as practical to community-run
```
evankanderson, line 57: It's not clear that this *and* the charter are needed. I would prefer to see this here as effectively a "bylaw" rather than as part of the charter (i.e. "the committee chooses to ..." rather than "the committee _is required to_ ...").

STEERING-COMMITTEE.md
```diff
@@ -77,12 +80,6 @@ options.
 Seats on the Steering Committee are held by an organization, not by the
 individual.
 
-The committee was created as the project was in its infancy, in order to tackle
-governance and overall project strategy. Because of the nature of the project
-and funding required, it was decided that strong corporate leadership was
-necessary for the project to ensure velocity. As the project grows and matures
-the KSC will, from time to time, consider if this policy should be changed.
-
```
evankanderson, line 85: This removal seems to foreclose the idea of later governance changes, which seems limiting.

STEERING-COMMITTEE.md
evankanderson: +1 -- Paul's proposal adds two seats for end-users to inform the community (via Steering) about how Knative is used in the world; this proposal would further reduce Steering's ability to hear feedback from the community (which feels like it's already a problem -- witness the 50+ people interested in the steering membership process with no current outlet).

Frankly, vendor seats have also been fairly disruptive to the community -- a huge amount of context has been lost in the last year as Google has shifted 100% of the seats assigned to it between different individuals, seemingly without a lot of handoff. This would be less of a problem if this wasn't *also* a majority bloc, but it feels like any progress that's been made over the last year on governance issues has been completely erased.

evankanderson: (Including the actual process of conducting a TOC election, which was removed from the current charter!)

STEERING-COMMITTEE.md
```diff
@@ -100,19 +97,25 @@ first name):
...
-Seats on the steering committee are allocated based upon contribution to the
-project by an organization. No final decision has been made on the exact
-formula.
+### Nomination of representatives
 
-As the project continues to grow, we expect to add additional seats to the
-committee, ensuring that those contributing to the project are properly
-represented.
+Seats on the steering committee are allocated based upon lifetime
+contributions to the project by an organization. The intention is
```
evankanderson, line 111: Lifetime contributions as a metric for representation feels like it is at odds with "being a 'truly open efffort'", because it makes it very difficult for a new contributing organization (or individual) to get a say in the policies that govern the organization. This method of measurement severely privileges the incumbent participants and *discourages* investment and engagement with the community.

I appreciate that it can feel unfair to have past contributions "age out" in terms of calculating representation, but this alternative seems overly conservative and against the interests of a healthy community.

If Google feels that there are *specific project issues* that need to be protected, it would be more productive to spell out those requirements and work on creating and empowering subcommittees to solve those problems and present them to Google for approval, rather than biasing the long-term governance in favor of older contributors.

evankanderson: In particular, play through this _hypothetical_ thought experiment:

> VMware contributes for another 6 months and ends up with (lifetime) 40k contributions as measured by devstats. VMware then decides that they are going to make an opposite investment, and stop investing in Knative, including maintenance. How long would it take for a replacement participant to "earn their way" to that seat starting from zero (no WG leads, ec) if it were left empty by VMware? 2 years? 3 years?

STEERING-COMMITTEE.md
```diff
@@ -100,19 +97,25 @@ first name):
...
+to ensure that those organizations are committed to the ongoing
+maintenance burden that adding to core implies, and express their
+commitment to the project and the project's users.
 
 - After a seat is allocated to an organization, the organization shall nominate
   a candidate to be confirmed by KSC. The committee reserves the right to not
-  confirm a candidate, in which the organization would need to nominate a new
+  confirm a candidate, in which case the organization would need to nominate a new
```
evankanderson, line 118: This feels a bit more slippery as a long-term solution with a structural vendor majority (observation).

Overall, this seems like a reasonable rule to prevent truly obnoxious behavior, but it's like bleach and ammonia -- two reasonable cleaners that are toxic when mixed.

STEERING-COMMITTEE.md
```diff
@@ -133,12 +136,15 @@ represented.
...
 
-The steering committee desires to always reach consensus.
+The steering committee desires to always reach consensus. The expectation
+is that lack of consensus requires multiple discussions and possibly public
+review, and that a vote would be an option of last resort.
 
-Decisions are made in meetings when a quorum of the members are present and may
-pass with majority of the committee supporting it.
+Decisions are made in meetings when a quorum of the members are present 
+and consensus is achieved, or in a meeting where all members are present
```
evankanderson, line 144: Is this "consensus of those present", or "consensus of all members (where those not present have already made their opinions known"?

STEERING-COMMITTEE.md
```diff
@@ -7,51 +7,55 @@ aliases:
...
-   administrators. This includes policy for the creation and administration of
-   [working groups](./working-groups/WORKING-GROUPS.md) and committees.
-1. Steward, control access, delegate access, and establishes processes
-   regarding, all Knative project resources and has the final say in the
-   disposition of those resources.
-1. Manage the Knative brand and decide which things can be called "Knative" and
-   how that mark can be used in relation to other efforts or vendors.
+1. Define, evolve, and promote the values of the project, and address any
+   [Code of Conduct](./CODE-OF-CONDUCT.md) issues that may arise, and
+   maintain confidentiality.
```
evankanderson, line 43: Combining "define the values of the project" with "maintain confidentiality" seems awkward. I think you want to separate the first and second sentences with a period, rather than ", and address"...

evankanderson (commented):
To be honest, I'd always seen the current governance structure as a "bootstrap" process to provide a decision-making and governance framework while these larger questions were being worked out (including "will this be a Google-majority project for all time"). Unfortunately, this proposal _in combination_ with the background discussion around Google's control of project trademark feels like a step backwards in terms of maturity to me -- the project is constrained both by Google's trademark ownership, _and_ intentionally by a set of restrictive governance structures.

I'm also uncomfortable with the way that this proposal seeks to bound the Steering Committee -- it seems like the Steering Committee is effectively prohibited from reviewing its own charter and adjusting it over time ("delegate all other responsibilities"), and no one outside the committee has a view as to the effectiveness of the committee because all meetings are private.

STEERING-COMMITTEE.md
duglin: :-) this is partially why I was pushing for changes to the voting structure in Paul's PR

devguyio:
As said on the public meeting, the community is giving Google a never-been-clearer signal of distrust in its unspoken intentions and future direction of the project, and I was hoping this proposal would provide a true sign of good faith. However, to me this proposal, given that it's only what's present in the proposal (Google having 3 lifetime seats out of a 5 seats total SC) seems cosmetic and I'm worried that it would be an insult to the community.

STEERING-COMMITTEE.md
maximilien: Additionally, having more representation would align better with missing #1: “empowers contributors”. As presented here lots of contributors would likely feel disenfranchised IMO.

STEERING-COMMITTEE.md
maximilien: Some type of dampening of “value” of contribution over time and even accelerating that dampening for contributors with lots of old contributions but less new ones could help. 

But then I imagine depending on dampening function, it will be a matter of playing around with the parameters to get to something fair and acceptable to all... I wonder if there is a precedence here in CNCF or other OSS organizations?

maximilien (changes requested):
Are there precedences for some of the items you are proposing, that corroborate and show heathy and vibrant communities when adopted?

STEERING-COMMITTEE.md
```diff
@@ -7,51 +7,55 @@ aliases:
   - /contributing/steering-committee/
 ---
 
-The Knative Steering Committee (KSC) is the ultimate authority for the Knative
-project, and governs all aspects of the project.
+The Knative Steering Committee (KSC) defines the governance for Knative
+in pursuit of the goal of an open, vibrant project with a productive
+community, a healthy ecosystem of vendors, and most of all satisfied
```
sebgoa, line 12: In general I think it is best to remove subjective terms from "charters". The definition of "vibrant" and "productive" is very subjective

STEERING-COMMITTEE.md
```diff
@@ -7,51 +7,55 @@ aliases:
...
+customers running their workloads on Knative clusters.
```
sebgoa, line 13: technically there are no customers of knative, there are users. There are customers of managed knative clusters and products based on knative. In general one cannot be a customer of an open source project since the community does not sell anything.

STEERING-COMMITTEE.md
```diff
@@ -7,51 +7,55 @@ aliases:
...
+  - [Nomination of representatives](#nomination-of-representatives)
 - [Decision process](#decision-process)
 - [Changes to the charter](#changes-to-the-charter)
 - [Getting in touch](#getting-in-touch)
 
+## Mission
+
+The Steering Committee's mission is to guide the project such that:
+
+1. Knative is a truly open effort which is inclusive and empowers contributors
```
sebgoa, line 35: I would remove truly because again that is subjective, otherwise you would have to define what a truly open effort is.

STEERING-COMMITTEE.md
sebgoa: I actually agree with the removal, I don't think this document is the right place for historical context. However it should lay out clearly how to amend the governance.

STEERING-COMMITTEE.md
```diff
@@ -100,19 +97,25 @@ first name):
 \* Open seats have been allocated to the organization per the committee rules,
   but have not been assigned to a member of the organization.
 
-### Allocation of seats
+### Composition
+
+As of 2020, the three organizations with the largest demonstrated 
+commitment to the project are Google, IBM/RedHat, and VMWare. For
+2021, a 5 person steering committee should be competent to
```
sebgoa, line 104: remove "should" use "is" or explain why it "should"

STEERING-COMMITTEE.md
```diff
@@ -100,19 +97,25 @@ first name):
...
+execute its responsibilities, with three seats for Google, one seat
+for IBM/RedHat, and one seat for VMWare.
```
sebgoa, line 106: I really don't get how you can have a "community" led projects and have a 100% corporate governance.

STEERING-COMMITTEE.md
sebgoa, line 113: this is really a downer for any independent contributor. You are thriving for a "vibrant" project but put the bar for membership in the steering at "corporation committed to ongoing maintenance burden", any freelancer/independent, small business, lone ranger etc is totally dis-incentivized to participate and engage fully. I understand technical leadership is in the TOC but some independent may want to engage and help steer as well. Saying "well you are not big enough, you don't have enough breadth to support maintenance" is an invitation for people to not engage.

We are now 2 years into this project and let's face it the contributions from startups and end-user companies are very very small, this is not a vibrant project.

STEERING-COMMITTEE.md
tpepper: The Kubernetes project looks for a certain number of contributions from the past year to determine how the community picks its steering body, eg: https://github.com/kubernetes/steering/blob/master/elections.md#eligibility-for-voting
One could imagine a variable number of minimum contributions required per year, with some non-zero and decreasing number of contributions in years N-2 and earlier allowing one a voice.

STEERING-COMMITTEE.md
```diff
@@ -7,51 +7,55 @@ aliases:
...
-   [working groups](./working-groups/WORKING-GROUPS.md) and committees.
-1. Steward, control access, delegate access, and establishes processes
-   regarding, all Knative project resources and has the final say in the
-   disposition of those resources.
-1. Manage the Knative brand and decide which things can be called "Knative" and
-   how that mark can be used in relation to other efforts or vendors.
+1. Define, evolve, and promote the values of the project, and address any
+   [Code of Conduct](./CODE-OF-CONDUCT.md) issues that may arise, and
+   maintain confidentiality.
+1. Review and approve changes to the scope of Knative core.
```
tpepper, line 44: This omits a key point on which I understood last week's meeting to hinge:  Veto.  The veto ability was described last week as something required for Google as the trademark holder.  But is not described in this PR.

Changes approved by contributors and approved by TOC and approved by steering is the easy path.

The concerning path for the community is the veto path:  Changes approved by contributors and approved by TOC but vetoed by one entity for reasons perhaps unclear to the community.  This scenario is a problem for many of us and is enabled by this PR.  Scenarios where Google would use this implicit veto power are not described in the PR.

Without that description, this PR does not grant the Steering Committee the power described in this line of the PR.  It may appear so, but when viewed in the context of comments in the meeting it does not.  This PR would establish that Google approval is required for changes to the scope of Knative core.  Steering as a whole would only "review" in an advisory position, not "review and approve" as noted.  Google is the one and only necessary approver of scope changes in Knative core, regardless of non-Google steering member input.

There was some discussion last week around if or how similar veto could be granted to other companies.  But that's the opposite of what the community desires.  The community wants representation and doesn't want vetos.  Lacking that it's hard to know if your investment as an individual or corporate developer in the project is going to be arbitrarily blocked by Google.  Any of us are accustomed to having our work judged on its merits and having to refactor if a group of our peers or senior reviewers deem our work insufficient.  In open source we are not accustomed to our work having the extra hurdle of needing review/approval by one corporation.

While OSI's definition is specific to license, I think many of us here are looking for similar attributes to those described in the open source definition but with respect to project governance, eg:
 * No Discrimination Against Persons or Groups
 * No Discrimination Against Fields of Endeavor
 * Must Not Be Specific to a Product
 * Must Not Restrict Other Software
 * Must Be Technology-Neutral
As specified in this PR, Google optionally can do the opposite on any of those points to its unique advantage and the community has no recourse _within_ the project.

STEERING-COMMITTEE.md
mattmoor: +1 (re: `truly`), but I don't think "open effort" is any more clear.  Open source is a spectrum and it's not clear where "effort" registers on that spectrum.

Perhaps @anicolao meant "open project", which would indicate more than simply "open _source_ project" (thus dropping the qualification "source") and would typically include "[open governance](http://opengovernance.dev/)", but that is not what this proposal outlines (#214 does, or at least gets us close)

rhuss (commented):
> Changes to the number of seats, which company seats are allocated to, and nominations to the committee are confirmed by majority vote of the committee members.

This sentence (albeit not changed by this PR) gets a new meaning with the change towards _lifetime contributions_ as the only metric for eligibility for holding a seat.

Afaiu the number of seats is dynamic, and depends on overall contributions to the project. But does this mean, if a new company would be eligible for a new seat (because of their contributions) and a new seat is added, that an additional seat for Google is added, too, to keep the majority (which is, if I interpret the proposal correctly, one of the driving forces in the proposal). In short: Is the number of seats always supposed to chagne pairwise, new vendor + google ? 

If this is the case, I think it would be fair to state in this proposal clearly that regardless what happens, the number of Google seats must have the majority. 

On the other hand, if this is not true, I would also say it clearly, that Google does not insist on keeping the majority of seats.

pmorie:
Apologies for my reply latency, I had to focus on a presentation yesterday that I gave earlier today on the Knative autoscaler. I'm reading this PR carefully now.

pmorie:
I must admit that I was expecting to read a PR that:

- recognized and affirmed the community alignment on community/214, with key points being:
  - a steering committee composed of individuals with their 'community hats' on
  - these individuals being chosen by the community in an election process
  - no majority of individuals on steering being employed by the same vendor
- expressed the compromise floated on thursday of last week in the public meeting:
  - drop end user seats from the proposal and hold size constant at 7 seats
  - mark holder keeps a reserved seat
  - google keeps a reserved 'founder' seat on steering
  - all other seats elected

...but verbal communication, especially in the heat of the moment, is imprecise. I'll grant the benefit of the doubt here and hope we can find something that works for all parties. In terms of this PR, there are some problematic bits:

- The ability of steering to define its own charter is removed
- Community-aligned elements of community/214 are not represented
- community/33 issue remains unaddressed:
  - steering seat allocation is no more clear
  - Google retains majority control over steering
- Lifetime contributions concept seems to provide odd incentives for ongoing maintenance

I continue puzzling over what the root issue is here. If the main concern for Google is scope and core compliance, why not delegate this to a compliance committee? Such a committee could have vendor representation and vote by consensus, which would seem to me to address that concern. Does that idea make sense? If so, could it work? Community members, could that work for you all, if that is the root issue?

pmorie:
One other thing, it would be good to have detail about the committees being delegated to by steering were created/composed/maintained.

dprotaso:
@anicolao Is https://github.com/knative/community/pull/277 intended to replace this PR?

anicolao:
Superseded by PR #277

PR #99, opened 2020-02-18
Add Dave Protasowski as Serving API Core WG technical lead.

mattmoor:
[Requirements](https://knative.dev/contributing/roles/#working-group-technical-lead):
- [X] **Recognized as having expertise in the group’s subject matter.**
  - Led the implementation of Conversion Webhook support for knative/serving.
  - Participated in the v1beta1/v1 API task force.
  - Many other features and improvements

- [X] **Approver for a relevant part of the codebase for at least 3 months.**
  - Approver for serving API since [June 2019](https://github.com/knative/serving/pull/4426)

- [X] **Member for at least 6 months.**
  - One of the earliest Knative (FKA Elafros) contributors!

- [X] **Primary reviewer for 20 substantial PRs.**
  - [47 L, XL, XXL in knative/serving](https://github.com/knative/serving/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Amerged+-author%3Adprotaso+reviewed-by%3Adprotaso+-label%3Asize%2FS+-label%3Asize%2FXS+-label%3Asize%2FM+)
  - [4 L, XL, XXL in knative/pkg](https://github.com/knative/pkg/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Amerged+-author%3Adprotaso+reviewed-by%3Adprotaso+-label%3Asize%2FS+-label%3Asize%2FXS+-label%3Asize%2FM+)

- [X] **Reviewed or merged at least 50 PRs.**
  - [67 reviewed](https://github.com/knative/serving/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Amerged+-author%3Adprotaso+reviewed-by%3Adprotaso+) + [58 merged](https://github.com/knative/serving/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Amerged+author%3Adprotaso) in serving
  - [10 reviewed](https://github.com/knative/pkg/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Amerged+-author%3Adprotaso+reviewed-by%3Adprotaso+) + [23 merged](https://github.com/knative/pkg/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Amerged+author%3Adprotaso) in pkg

- [X] Sponsored by the technical oversight committee.

/cc vaikas evankanderson

> NOTE: This also removes @dgerd who announced last week that he plans to step down :(

PR #287, opened 2020-09-22
Removing Sam as Doc WG lead

RichieEscarez:
Sam stopped working on the project some time in March.

### DIFF
--- a/working-groups/WORKING-GROUPS.md
+++ b/working-groups/WORKING-GROUPS.md
@@ -119,7 +119,6 @@ Knative documentation, especially the [Docs](https://github.com/knative/docs/) r
 
 | &nbsp;                                                   | Leads      | Company | Profile                                 |
 | -------------------------------------------------------- | ---------- | ------- | --------------------------------------- |
-| <img width="30px" src="https://github.com/richieescarez.png"> | Richie Escarez | Google  | [richieescarez](https://github.com/richieescarez) |
 | <img width="30px" src="https://github.com/abrennan89.png"> | Ashleigh Brennan | Red Hat  | [abrennan89](https://github.com/abrennan89) |
 
 ## Eventing
--- a/working-groups/WORKING-GROUPS.md
+++ b/working-groups/WORKING-GROUPS.md
@@ -119,7 +119,7 @@ Knative documentation, especially the [Docs](https://github.com/knative/docs/) r
 
 | &nbsp;                                                   | Leads      | Company | Profile                                 |
 | -------------------------------------------------------- | ---------- | ------- | --------------------------------------- |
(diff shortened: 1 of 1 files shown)


RichieEscarez:
assign/ @abrennan89

mattmoor:
Hey @RichieEscarez, welcome back.  Sorry to hear that @samodell might not come back, I hope she's doing better.

The process for new working group leads is spelled out [here](https://github.com/knative/community/blob/master/ROLES.md#role-summary), and the PRs adding new leads are generally asked to follow a form like [this](https://github.com/knative/community/pull/99).  I think the idea of "replacing" WG leads sets a bad precedent, so I'd prefer if we removed Sam separately from the process above.

I don't think the quantitative aspects of our WG lead "template" cover  everything though, especially prolonged absence.  I think I'd personally be more comfortable if the bulk of the contributions justifying WG leadership had a bias towards recent contribution because in that role the importance of context is fairly high.

RichieEscarez:
Thanks Matt! Sam is definitely off the project. She is back to being "Sam" but decided to move back into working on GCP products rather than OSS.

Sounds good (to just removing Sam in this PR). My motivation here had two parts. I added this to todays Docs WG meeting with the intention to discuss this there but given the lack of sessions and attendance (today and through Aug), I knew that I would/could get input/feedback this route.

Question: Given the two lead definitions, my goal was to try and help fulfill the expectations listed. I know much as changed since the beginning of the year especially in the Docs WG and many formal WG processes seem not to exist or are lacking an owner. That type of support is not articulated in the community guidelines so I'm wondering if it would be considered?

Also, regarding "absence", that's not spelled out either? I did ramp down significantly but was still an active member over Q2 and Q3. That said, the scope of time for someones contributions are also lacking in the community guidelines (does that mean I still can check all the boxes)? Anyway, ill try again and bring up whether I should fill out the Docs WG lead from in the next Docs meeting.

evankanderson:
Welcome back @RichieEscarez !

I agree with Matt that making this two PRs will be clearer, and I'm sorry to hear that we won't be seeing Sam again.

Given your previous time with the project, I'd expect that you'd probably be quickly contributing effectively as a WG lead -- I'd defer decision on the nomination to @abrennan89 , who has been trying to right the process ship that slipped during the time when there was no WG lead.

I'd advise her to sit down with you and make sure that you two have the same vision and goals (possibly by working together for at least a few weeks) before adding a second WG lead. Having an extra set of hands to bail out a boat can be helpful, unless you each end up bailing into each others' boats (i.e. working at cross-purposes). Working together for a few weeks and planning out what the next few months look like should help both of you figure out if it's a good fit, since I know Ashleigh has a few opinions about what her ideal stack would look like.

evankanderson (commented):
/lgtm
/approve

Since this is a pure removal.

I can follow up with a PR for peribolos.

PR #294, opened 2020-09-25
Fix peribolos config file for Knative by removing jessiezcc

chizhg:
The error is `Configuration failed: failed to configure knative members: all team members/maintainers must also be org members: jessiezcc`, see https://prow.knative.dev/view/gs/knative-prow/logs/ci-knative-peribolos/1309538632675102723#1:build-log.txt%3A8

/cc @mattmoor


mattmoor (commented):
/lgtm
/approve

PR #303, opened 2020-09-30
add kn-diag repo into peribolos/knative-sandbox.yaml

cdlliuy:


tcnghia:
@cdlliuy can you please resolve the conflict so the tests can run? thanks

Issue #137, opened 2020-05-19
Sort out knative.team


evankanderson:
Ryan got us a GSuite org. We need to do the following:

* [x] Add new TOC / Steering members to the org
* [x] Get calendar sharing working
* [x] Get team drive working
* [ ] Document setup
* [ ] Update documentation

evankanderson:
Set up time on 6/16 to do this.

mattmoor:
@evankanderson is this a dupe with: https://github.com/knative/community/issues/86

evankanderson:
I think this was part of #86. This was about having the GSuite configuration working to share; I'm going to use #86 for doing the actual cutover.

evankanderson:
/close

Duplicate of #86

Issue #86, opened 2020-02-10
Move to Knative Community owned GSuite account


bsnchan:
## Current State

The Knative GSuite tools use the `google.com` domain at the moment. This affects:
* Google Drive
  * Folks outside of Google can't control the ownership of the community drive itself
* Knative Community Calendar
  * Folks outside of Google have read/write access but don't have admin privileges
* Hangouts
  * We still use Hangouts for a lot of WG meetings which require someone from Google to let folks into the meeting
* Google Groups
  * I believe all the knative-* google groups are tied to the `google.com` GSuite org as well

## Proposal

Create a knative.dev GSuite org for the community and shift all the community owned artifacts to this org.

## Steps

* [x] Add new TOC / Steering members to the org
* [x] Get calendar sharing working
* [x] Get team drive working
* [x] Document setup
* [ ] Update documentation

bsnchan:
Hey @rgregg @isdal - this feels like an easy one to tackle. Is there any chance we can get some budget for this?

mattmoor:
tl;dr I'd like to plug for an increase in the priority of this.

I had a quick sync with @dims and having a community Gsuite org seems to be one of the cornerstones of how they are migrating the pieces of the K8s CI/CD/Release infrastructure from Google-only, to community based.

It sounds like they have significant automation for managing the Gsuite org (see [here](https://github.com/kubernetes/k8s.io/blob/master/groups/groups.yaml)), and by the end of our chat I was in a very "shut up and take my money" mood.

That this is being used to address the exact-same problems on the Kubernetes side sets a very strong precedent that we didn't have previously motivating this, and the ways they are using it (which we should follow) make this blocking for other items on this board.

rgregg:
I'm working on setting up a G-Suite account for us to use.

rgregg:
We have a G-Suite account now, using `knative.team`. I believe @evankanderson is working on the migration details for our membership groups and shared drive.

evankanderson:
/assign @evankanderson 

We will be moving Team Drive & Calendar on the weekend of July 4 to attempt to minimize disruption.

mattmoor:
@evankanderson You mentioned sitting down with Tomas last week to work on this, any update?

evankanderson:
Update is:

The following calendar and document should be publicly visible:
[...]

PR #307, opened 2020-10-02
Initial work on gdocs migration tool with tcnghia.

evankanderson:
/assign @tcnghia 

I dropped a brief plan into the directory with pointers to the GSuite API. We'll also need to figure out how to set up credentials (I don't care whether this is a shared project, or we each have our own, but Google is somewhat twitchy  about apps getting access to `https://www.googleapis.com/auth/drive.file` or `https://www.googleapis.com/auth/drive` permissions, one of which we will need.


tcnghia:
Thanks @evankanderson 
/lgtm

evankanderson:
A few notes on this; I had a 1:1 meeting with Nghia, and wanted to put the plan down more formally.

This helps to solve #137 and #86. Based on the Google.com current sharing settings, people can't move docs out of the current drive (because it's owned by Google). "Make a copy" does work, but there are a lot of docs, and some of them are actually in-flight. So we're going to automate this, make a copy of all the docs, and then put a banner on the current docs pointing to the new location and change the team drive settings so that the new drive is shared with knative.dev and the old drive is limited to view permissions.

We're checking in the tool we use, because it makes sense to review and archive the tool, but we're not vendoring all the go modules or using the root of the repo because this will hopefully be a one-off, and we may move it to an "automation" repo if one exists.

Note that the new drive will allow for "share to web", which should make publishing docs a bit easier, so I'm a little motivated every time I get a GDocs sharing request.

PR #332, opened 2020-10-21
Load all files metadata into a forest

tcnghia:
* Increase page size for faster operation
* Store all File metadata in a forest

Next step would be to copy them from one drive to the other.

/assign @evankanderson

### DIFF
--- a/mechanics/tools/gdocs-migrate/forest.go
+++ b/mechanics/tools/gdocs-migrate/forest.go
@@ -1,0 +1,63 @@
+package main
+
+type File struct {
+	id       string
+	name     string
+	parentID string
+	isFolder bool
+}
+
+type Forest struct {
+	fileByID map[string]File
+	children map[string][]File
+	leaves   []File
+}
+
+func NewForest() Forest {
+	return Forest{
+		fileByID: make(map[string]File),
+		children: make(map[string][]File),
+	}
+}
+
+func (fo *Forest) Add(f File) {
+	fo.fileByID[f.id] = f
+
+	fo.children[f.parentID] = append(fo.children[f.parentID], f)
+	if !f.isFolder {
+		fo.leaves = append(fo.leaves, f)
+	}
+}
+
+func (fo *Forest) GetFiles() []File {
+	return fo.leaves
+}
+
+func (fo *Forest) Size() int {
+	return len(fo.fileByID)
+}
+
+func (fo *Forest) FolderCount() int {
+	return fo.Size() - len(fo.GetFiles())
+}
+
+func (fo *Forest) GetPath(f File) []string {
+	parents := []string{}
+	cur := f
+	p, hasParent := fo.fileByID[cur.parentID]
+	for hasParent {
+		if p.name != "" {
+			parents = append(parents, p.name)
+		}
+		cur = p
+		p, hasParent = fo.fileByID[cur.parentID]
+	}
+	return reverse(parents)
+}
+
+func reverse(s []string) []string {
+	for i, j := 0, len(s)-1; i < j; i, j = i+1, j-1 {
+		s[i], s[j] = s[j], s[i]
+	}
+	return s
+}
--- a/mechanics/tools/gdocs-migrate/forest.go
+++ b/mechanics/tools/gdocs-migrate/forest.go
@@ -44,13 +44,9 @@ func (fo *Forest) FolderCount() int {
 func (fo *Forest) GetPath(f File) []string {
 	parents := []string{}
 	cur := f
-	p, hasParent := fo.fileByID[cur.parentID]
-	for hasParent {
-		if p.name != "" {
-			parents = append(parents, p.name)
-		}
-		cur = p
-		p, hasParent = fo.fileByID[cur.parentID]
+	for cur.parentID != "" {
+		cur = fo.fileByID[cur.parentID]
+		parents = append(parents, cur.name)
 	}
 	return reverse(parents)
 }
--- a/mechanics/tools/gdocs-migrate/forest.go
+++ b/mechanics/tools/gdocs-migrate/forest.go
@@ -41,6 +41,11 @@ func (fo *Forest) FolderCount() int {
 	return fo.Size() - len(fo.GetFiles())
 }
 
+// GetPath returns all the ancestor of a given File in decreasing
+// seniority.
+//
+// This should only be called after the whole forest has ben loaded,
+// i.e. Add(...) has been called for all the Files.
 func (fo *Forest) GetPath(f File) []string {
 	parents := []string{}
 	cur := f
--- a/mechanics/tools/gdocs-migrate/main.go
+++ b/mechanics/tools/gdocs-migrate/main.go
@@ -3,13 +3,15 @@ package main
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"google.golang.org/api/drive/v3"
 )
 
 const (
-	KnativeCommunityDriveId = "0APnJ_hRs30R2Uk9PVA"
-	KnativeTeamDriveId      = "0AM-QGZJ-HUA8Uk9PVA"
+	KnativeCommunityDriveID = "0APnJ_hRs30R2Uk9PVA"
+	KnativeTeamDriveID      = "0AM-QGZJ-HUA8Uk9PVA"
+	FolderMimeType          = "application/vnd.google-apps.folder"
 )
 
 func main() {
@@ -26,28 +28,49 @@ func main() {
 	}
 
 	drives := map[string]string{
-		"Knative Community [old]": KnativeCommunityDriveId,
-		"Knative Team":            KnativeTeamDriveId,
+		"Knative Community Drive": KnativeCommunityDriveID,
+		"Knative Team Drive":      KnativeTeamDriveID,
 	}
 
+	forest := NewForest()
 	for name, id := range drives {
-		// List files from Knative Community Drive.
-		r, err := srv.Files.List().PageSize(10).
-			Corpora("drive").
-			IncludeItemsFromAllDrives(true).
-			SupportsAllDrives(true).
-			DriveId(id).
-			Fields("nextPageToken, files(id, name)").Do()
-		if err != nil {
-			log.Fatal("Unable to retrieve files:", err)
-		}
-		fmt.Printf("---------------------------\n%s files:\n---------------------------\n", name)
-		if len(r.Files) == 0 {
-			fmt.Println("No files found.")
-		} else {
-			for _, i := range r.Files {
-				fmt.Printf("%s (%s)\n", i.Name, i.Id)
+		fmt.Println("Loading directory from ... ", name)
+		forest.Add(File{
+			id:       id,
+			name:     name,
+			isFolder: true,
+		})
+		var pageToken *string
+		for pageToken == nil || *pageToken != "" {
+			// List files from Knative Community Drive.
+			request := srv.Files.List().PageSize(1000).
+				Corpora("drive").
+				IncludeItemsFromAllDrives(true).
+				SupportsAllDrives(true).
+				DriveId(id).
+				Fields("nextPageToken, files(id, name, parents, mimeType)").
+				OrderBy("folder,modifiedTime desc,name")
+			if pageToken != nil {
+				request.PageToken(*pageToken)
 			}
+			r, err := request.Do()
+			if err != nil {
+				log.Fatal("Unable to retrieve files:", err)
+			}
+			pageToken = &r.NextPageToken
+			for _, i := range r.Files {
+				forest.Add(File{
+					id:       i.Id,
+					name:     i.Name,
+					parentID: i.Parents[0],
+					isFolder: i.MimeType == FolderMimeType,
+				})
+			}
+			fmt.Printf("Loaded %d files, %d folders\n", len(forest.GetFiles()), forest.FolderCount())
 		}
+	}
+
+	for _, f := range forest.GetFiles() {
+		fmt.Printf("[%s] %s\n", strings.Join(forest.GetPath(f), "/"), f.name)
 	}
 }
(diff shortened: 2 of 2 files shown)


mechanics/tools/gdocs-migrate/forest.go
```diff
@@ -0,0 +1,60 @@
...
+
+type File struct {
+	id       string
+	name     string
+	parentId string
+	isFolder bool
+}
+
+type Forest struct {
+	self     map[string]File
```
vagababov, line 11: why self is a map? one would expect self is a single entry.

mechanics/tools/gdocs-migrate/forest.go
```diff
@@ -0,0 +1,60 @@
...
+func (fo *Forest) GetFiles() []File {
+	return fo.leaves
+}
+
+func (fo *Forest) Size() int {
+	return len(fo.self)
+}
+
+func (fo *Forest) GetPath(f File) []string {
+	parents := []string{}
```
vagababov, line 41: in theory you can just store it in the file itself when you create the file, then it's just `return path`

mechanics/tools/gdocs-migrate/forest.go
tcnghia: I'll rename to `fileById`

mechanics/tools/gdocs-migrate/forest.go
tcnghia: we don't have a guarantee that we receive the parent metadata before the child.

mechanics/tools/gdocs-migrate/forest.go
tcnghia: Done

mechanics/tools/gdocs-migrate/forest.go
```diff
@@ -0,0 +1,59 @@
...
+
+type File struct {
+	id       string
+	name     string
+	parentId string
+	isFolder bool
+}
+
+type Forest struct {
+	fileById map[string]File
```
vagababov, line 11:
```suggestion
	fileByID map[string]File
```

mechanics/tools/gdocs-migrate/forest.go
```diff
@@ -0,0 +1,59 @@
...
+	children map[string][]File
+	leaves   []File
+}
+
+func NewForest() Forest {
+	return Forest{
+		fileById: make(map[string]File),
```
vagababov, line 18:
```suggestion
		fileByID: make(map[string]File),
```

mechanics/tools/gdocs-migrate/forest.go
vagababov: How do you build path if you don't have parent MD?

mechanics/tools/gdocs-migrate/main.go
```diff
@@ -3,13 +3,15 @@ package main
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"google.golang.org/api/drive/v3"
 )
 
 const (
 	KnativeCommunityDriveId = "0APnJ_hRs30R2Uk9PVA"
```
vagababov, line 12:
```suggestion
	KnativeCommunityDriveID = "0APnJ_hRs30R2Uk9PVA"
```

mechanics/tools/gdocs-migrate/main.go
```diff
@@ -3,13 +3,15 @@ package main
...
 	KnativeTeamDriveId      = "0AM-QGZJ-HUA8Uk9PVA"
```
vagababov, line 13:
```suggestion
	KnativeTeamDriveID      = "0AM-QGZJ-HUA8Uk9PVA"
```

mechanics/tools/gdocs-migrate/main.go
```diff
@@ -26,28 +28,51 @@ func main() {
...
+				SupportsAllDrives(true).
+				DriveId(id).
+				Fields("nextPageToken, files(id, name, parents, mimeType)").
+				OrderBy("folder,modifiedTime desc,name")
+			if pageToken != nil {
+				request.PageToken(*pageToken)
+			}
+			r, err := request.Do()
+			if err != nil {
+				log.Fatal("Unable to retrieve files:", err)
```
vagababov, line 58: perhaps add debug info, like request?

mechanics/tools/gdocs-migrate/main.go
```diff
@@ -26,28 +28,51 @@ func main() {
...
-				fmt.Printf("%s (%s)\n", i.Name, i.Id)
+				forest.Add(File{
+					id:       i.Id,
+					name:     i.Name,
+					parentId: i.Parents[0],
+					isFolder: i.MimeType == FolderMimeType,
+				})
 			}
+			fileCount := len(forest.GetFiles())
+			folderCount := forest.Size() - fileCount
```
vagababov, line 70: this feels like a function on forest.

tcnghia: Done

mechanics/tools/gdocs-migrate/main.go
```diff
@@ -26,28 +28,49 @@ func main() {
 	}
 
 	drives := map[string]string{
-		"Knative Community [old]": KnativeCommunityDriveId,
-		"Knative Team":            KnativeTeamDriveId,
+		"Knative Community Drive": KnativeCommunityDriveId,
+		"Knative Team Drive":      KnativeTeamDriveId,
```
vagababov, line 32: well those also have to be changed 😁

vagababov (commented):
/lgtm
/hold
for @evankanderson

mechanics/tools/gdocs-migrate/forest.go
```diff
@@ -0,0 +1,63 @@
...
+		fileByID: make(map[string]File),
+		children: make(map[string][]File),
+	}
+}
+
+func (fo *Forest) Add(f File) {
+	fo.fileByID[f.id] = f
+
+	fo.children[f.parentID] = append(fo.children[f.parentID], f)
+	if !f.isFolder {
```
evankanderson, line 27: I think this is fine; another way to do this would be to put files into either `fileByID` or `children` based on whether or not they are a folder. It would probably require a slightly more complex structure (`struct {File, [] File}`) instead of the current slice.

mechanics/tools/gdocs-migrate/forest.go
evankanderson: `GetPath` is only called after all the folder data is collected, so we'll have the data then, but we won't necessarily have it during `Add`.

mechanics/tools/gdocs-migrate/forest.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (fo *Forest) FolderCount() int {
+	return fo.Size() - len(fo.GetFiles())
+}
+
+func (fo *Forest) GetPath(f File) []string {
+	parents := []string{}
+	cur := f
+	p, hasParent := fo.fileByID[cur.parentID]
+	for hasParent {
```
evankanderson, line 48: I think this could be `for cur.parentID != ""` if we don't get an entity for the root, but all others have actual IDs.

mechanics/tools/gdocs-migrate/forest.go
```diff
@@ -0,0 +1,63 @@
...
+		if p.name != "" {
+			parents = append(parents, p.name)
```
evankanderson, line 50: Should this be a panic or something, rather than skipping over parents that don't have names?

mechanics/tools/gdocs-migrate/forest.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+		cur = p
+		p, hasParent = fo.fileByID[cur.parentID]
+	}
+	return reverse(parents)
```
evankanderson, line 55: Can this be [`sort.Reverse`](https://golang.org/pkg/sort/#Reverse) ?

mechanics/tools/gdocs-migrate/main.go
```diff
@@ -3,13 +3,15 @@ package main
...
 
 	"google.golang.org/api/drive/v3"
 )
 
 const (
-	KnativeCommunityDriveId = "0APnJ_hRs30R2Uk9PVA"
-	KnativeTeamDriveId      = "0AM-QGZJ-HUA8Uk9PVA"
+	KnativeCommunityDriveID = "0APnJ_hRs30R2Uk9PVA"
+	KnativeTeamDriveID      = "0AM-QGZJ-HUA8Uk9PVA"
+	FolderMimeType          = "application/vnd.google-apps.folder"
```
evankanderson, line 14: I may add a flag that would be a doc ID prefix to allow us to test this without hitting the whole folder. WDYT?

mechanics/tools/gdocs-migrate/main.go
```diff
@@ -26,28 +28,49 @@ func main() {
...
-				fmt.Printf("%s (%s)\n", i.Name, i.Id)
+				forest.Add(File{
+					id:       i.Id,
+					name:     i.Name,
+					parentID: i.Parents[0],
+					isFolder: i.MimeType == FolderMimeType,
+				})
 			}
+			fmt.Printf("Loaded %d files, %d folders\n", len(forest.GetFiles()), forest.FolderCount())
 		}
```
evankanderson, line 70: Extract to function?

mechanics/tools/gdocs-migrate/main.go
```diff
@@ -26,28 +28,49 @@ func main() {
...
-			fmt.Println("No files found.")
-		} else {
+		fmt.Println("Loading directory from ... ", name)
+		forest.Add(File{
+			id:       id,
+			name:     name,
+			isFolder: true,
+		})
+		var pageToken *string
+		for pageToken == nil || *pageToken != "" {
```
evankanderson, line 44: If you're handling `pageToken == ""`, any reason pageToken needs to be a `*string` and not a `string`?

mechanics/tools/gdocs-migrate/forest.go
evankanderson: But Victor's point is that we're only passing in `i.Parents[0]`, but the API will give us all the ancestors.

mechanics/tools/gdocs-migrate/main.go
```diff
@@ -26,28 +28,49 @@ func main() {
...
 			for _, i := range r.Files {
-				fmt.Printf("%s (%s)\n", i.Name, i.Id)
+				forest.Add(File{
+					id:       i.Id,
+					name:     i.Name,
+					parentID: i.Parents[0],
+					isFolder: i.MimeType == FolderMimeType,
+				})
 			}
+			fmt.Printf("Loaded %d files, %d folders\n", len(forest.GetFiles()), forest.FolderCount())
```
evankanderson, line 69: I'm assuming this works as a sort of progress bar when inside this loop?

evankanderson (commented):
A bunch of minor comments, but feel free to submit and then adjust in the future if you like the suggestions.

(Cancel the hold if you just want to submit.)

mechanics/tools/gdocs-migrate/forest.go
tcnghia: I'll keep it this way for now and see where this leads us, and will change the interface/data as needed.

mechanics/tools/gdocs-migrate/forest.go
tcnghia: that require sort.Interface, unfortunately...

mechanics/tools/gdocs-migrate/forest.go
tcnghia: the Parents array give us all the direct Parents (apparently one File may belong to a few Parent -- fortunately none of our File is that way).

that also means Parents[0] is all that we get... 

I've updated doc that we should call these after all data has been loaded.

mechanics/tools/gdocs-migrate/main.go
tcnghia: That sounds good to me. But this runs really fast and so far I've not seen 429 🤘

mechanics/tools/gdocs-migrate/forest.go
tcnghia: That's a good idea. Done.

mechanics/tools/gdocs-migrate/forest.go
tcnghia: It's leftover -- I've updated the code.

mechanics/tools/gdocs-migrate/forest.go
```diff
@@ -0,0 +1,64 @@
...
+func (fo *Forest) FolderCount() int {
+	return fo.Size() - len(fo.GetFiles())
+}
+
+// GetPath returns all the ancestor of a given File in decreasing
+// seniority.
+//
+// This should only be called after the whole forest has ben loaded,
+// i.e. Add(...) has been called for all the Files.
+func (fo *Forest) GetPath(f File) []string {
```
vagababov, line 49: This can also be cached if in the struct if it's going to be called more than once.

mechanics/tools/gdocs-migrate/main.go
tcnghia: Done

mechanics/tools/gdocs-migrate/main.go
tcnghia: I use `nil` for first page, and "" for last page.

mechanics/tools/gdocs-migrate/main.go
tcnghia: Yes, exactly..

mechanics/tools/gdocs-migrate/forest.go
tcnghia: Sure thing, if it is going to be a bottle neck I'll do that.

mechanics/tools/gdocs-migrate/forest.go
evankanderson: I think using sort.StringSlice would do that for you, but this is fine, too.

PR #338, opened 2020-10-23
fix the Productivity wg meeting time

n3wscott:
We moved to weekly to include talk for the knative testing frameworks.


mattmoor (approved):
/lgtm

PR #344, opened 2020-10-27
Sc 2020 election -- fix broken links

jberkus:
This PR fixes three broken relative links in SC/Readme and TOC/Readme.

I changed them to absolute links to avoid this kind of breakage in the future.  I'll suggest that we should adopt a pragma of not using relative links that traverse directory trees, becuase they're really hard to maintain correctly.

Assign: @bsnchan @pmorie


bsnchan:
Thanks for this!
/lgtm
/approve

PR #357, opened 2020-11-05
Nominate vaikas for SC

mattmoor:
/hold
/assign @vaikas 

Ville, LMK when you have a chance to give some real words for why you want to run, but in the meantime I went with parody to get something posted before the deadline. 😉

PR #367, opened 2020-11-10
vaikas 2020 SC nomination

vaikas:
Follow onto #357 that may or may not have more context on the why.


jberkus:
This PR fulfills all requirements for an candidate for the election.

@bsnchan, please merge.

Issue #9, opened 2019-05-30
Add docs around use of name and logo


duglin:
I couldn't find any docs that describe how/when people can use the Knative name and logo. We should probably say something about this.

For inspiration, see:
https://github.com/kubernetes/kubernetes/blob/master/logo/usage_guidelines.md
and
https://www.linuxfoundation.org/trademark-usage/

/cc @rgregg

duglin:
@knative/knative-steering-committee

rgregg:
I'm working on this.

duglin:
Any update on this one?

duglin:
I know there's some talk going on about this, but I wanted to raise this issue back to the top of the SC's todo list. We really need a definitive statement about this on our website so that there is a clear, concise and consistent set of rules that everyone knows they need to follow. There should not need to be off-line discussions just to know what they are. Off-line chats should only be necessary if someone has follow-on questions/concerns.

Can we please get a target date and an owner for this so that it doesn't linger?

rgregg:
I'm assigned and working on this. As with anything involving lawyers, progress is slow. @duglin I'm looking to gather more requirements around what the community wants the guidelines to be. Are you the right contact from IBM to connect with so we can understand the asks from IBM?

duglin:
@rgregg sure

rgregg:
@thisisnotapril is taking on this work.

duglin:
@thisisnotapril can we get an ETA on this? It's been almost a year since the initial request.

thisisnotapril:
@duglin should have it this coming week!

duglin:
@thisisnotapril any update?

thisisnotapril:
@duglin thanks for your patience - coming very soon! I am having other Steering members review a commitment; then will add the PR.

PR #147, opened 2020-06-08
Create BRANDING.MD

thisisnotapril:
Guidelines for usage of Knative logo

PR #388, opened 2020-11-30
Create BRANDING.MD

duglin:
Guidelines for usage of Knative logo and trademark

Replaces #147 

Fixes: #9 

### DIFF
--- a/BRANDING.MD
+++ b/BRANDING.MD
@@ -1,0 +1,39 @@
+## Knative Branding Guidelines
+
+As a general rule, Knative follows the [Linux Foundation trademark
+usage](https://www.linuxfoundation.org/trademark-usage/) rules. As a high-level
+summary, the following is made for easy reference.
+
+The Knative logo and trademark is made available here with the following
+guidelines:
+
+ **You may:**
+ - use them for non-commercial uses such as t-shirts and stickers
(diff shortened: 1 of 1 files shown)


bsnchan:
/lgtm

@ronavn - can you double check to make sure this is okay with the Google Legal team?

BRANDING.MD
```diff
@@ -0,0 +1,39 @@
+## Knative Branding Guidelines
+
+As a general rule, Knative follows the [Linux Foundation trademark
+usage](https://www.linuxfoundation.org/trademark-usage/) rules. As a high-level
+summary, the following is made for easy reference.
```
evankanderson, line 5: This last bit is confusing -- if there's a conflict between the text here and the Linux Foundation usage rules, which set of rules governs?

(Note that this could happen by Linux Foundation updating their rules.)

duglin: @evankanderson updated it - see if that addresses your concen

thisisnotapril: Do we need to say this bit even? If the summary as written covers the majority of the uses cases, and anything outside that would go to the TC, shouldn't that be enough? Referencing another guideline that we don't have any say in could just be confusing.

PR #400, opened 2020-12-04
Updates steering committee's email group

bsnchan:
# Changes

Updates steering committee's email group. We decided to move to the knative.team Google group so we can self manage it :)


evankanderson (commented):
/lgtm

Issue #101, opened 2020-02-25
Reconcile TBD Google tech writers


mattmoor:
I created this temporary team with elevated access to be resolved by steering: https://github.com/orgs/knative/teams/tbd-google-tech-writers/members

vaikas:
Given that @abrennan89 is the Docs WG lead, want to make sure she's looped in.
I'm honestly not clear what needs to happen here based on the description.
I see the following group:
https://github.com/orgs/knative/teams/docs-writers/members?query=membership%3Achild-team

So, I see that @abrennan89 is a member because of this:
https://github.com/orgs/knative/teams/docs-wg-leads

And the other 3 are part of the tbd-google-tech-writers.
Just trying to understand what makes the Google part necessary and not a member of say docs-writers. If somebody could clarify, then I might be able to do something about this :)

@ellenevans @RichieEscarez @carieshmarie Can you let me know what prevents us from moving to the docs-writers?

Thanks!

mattmoor:
Moving to docs-writers as what role?

abrennan89:
I'm not sure what all of this is about, but I see that Ellen and Caroline are listed in those groups. AFAIK they're not involved in Knative anymore. Richie still is but spends very limited time working on this.

vaikas:
Yeah, I just didn't have enough context from the issue description. There's a few of these issues that were created and I'm just trying to go through the backlog of serving issues and try to make headway. The way I saw it that there was couple of things at play:
 1. A group that @mattmoor created that we want to get rid off
 2. Is there a reason for this group or can the folks in that group me moved to some existing group
 3. If no, what group. If yes, why.

mattmoor:
Ack, I was sort of expecting these issues to be handled while context was a bit fresher. 😉 

The initial ACL restructure was intended to map ACLs to project roles, and these issues were opened for every deviation where folks were even mildly hesitant about losing access.

RichieEscarez:
Hey all! I'm honestly not sure what these are or how the ACLs work these days.

For context, @carieshmarie was added in the beginning to make sure someone had the privileges to cover to Sam and I. I chatted with her recently and she doesn't feel the need to keep her permissions if that means we align with the Knative org rules.  @ellenevans stepped in temporarily to help keep docs efforts going when Sam switched projects and I was away (temporarily loaned out to a priority project and then out-of-office for Carers leave).

[...]

PR #430, opened 2021-02-03
address issue 101

vaikas:
As per #101 add @RichieEscarez into the Release Leads so that we can remove the temporary group.


markusthoemmes (approved):
/lgtm
/approve

Issue #245, opened 2020-08-28
Revisit: K8s minimum version principle


mattmoor:
See here: https://kubernetes.io/blog/2020/08/26/kubernetes-release-1.19-accentuate-the-paw-sitive/#increase-kubernetes-support-window-to-one-year

cc @dprotaso @pmorie

mattmoor:
@dprotaso do you want to take a crack at this now that you're back? 🥺

dprotaso:
yeah it's on my backlog

evankanderson:
(ping)

dprotaso:
related: https://github.com/kubernetes/sig-release/issues/1290

dprotaso:
And https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md

dprotaso:
woo structured data https://github.com/kubernetes/sig-release/blob/master/releases/schedule.yaml

dprotaso:
### [Recap: K8s minimum version principle](https://github.com/knative/community/blob/master/mechanics/RELEASE-VERSIONING-PRINCIPLES.md#k8s-minimum-version-principle)

> We (the community) support the range of upstream minor Kubernetes versions that are Kubernetes community supported at the time of the cut. Only the latest patch release of the minor version is tested for support. Support for versions prior to that are best effort, or by vendors.
> 
> Adoption of newer K8s minimum versions will also be on a best-effort basis. When a Kubernetes version is released it's availability in downstream distributions is often lagging. When these delays affect our planned min K8s version we'll update the version table below and make it known in the release's notes.
> 

### Upstream EOL Changes
With v1.19 the support window is now a year. This change has been rolled back into v1.17 & v1.18 (see [PR](https://github.com/kubernetes/sig-release/commit/27f1b9f155d23475d6e99b694c061887814baf81)). 

### Current Version Pairing between K8s & Knative

Knative | Min K8s | notes
-|-|-
v0.19 | v1.17 |
v0.20 | v1.17 | EOLs 2 weeks after
v0.21 | v1.18 |
v0.22 | v1.18 | EOLs 2 weeks after
v0.24 | v1.19 | 
v0.25 | v1.19 |
v0.26 | v1.19 | EOLs 1 week after
v0.27 | v1.20 |

[...]

PR #459, opened 2021-02-16
extend to .23, remove v1beta1 eventing in .23

vaikas:
Add the next two releases .22 and .23 to the calendar. As per https://github.com/knative/eventing/issues/4836 also mark the eventing v1beta1 APIs as removed then.

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/mechanics/RELEASE-VERSIONING-PRINCIPLES.md
+++ b/mechanics/RELEASE-VERSIONING-PRINCIPLES.md
@@ -226,7 +226,43 @@ tuple
    <td><strong>Availability</strong>
    </td>
   </tr>
-    <tr>
+   <tr>
+   <td>0.23.x
+   </td>
+   <td>2021-05-18
+   </td>
+   <td>2021-11-02
+   </td>
+   <td>v1.18.x
+   </td>
+   <td>v1
+   </td>
+   <td>v1
+   </td>
+   <td>v1
+   </td>
+   <td>default
+   </td>
+  </tr>
+   <tr>
+   <td>0.22.x
+   </td>
+   <td>2021-04-06
+   </td>
+   <td>2021-09-21
+   </td>
+   <td>v1.18.x
+   </td>
+   <td>v1
+   </td>
+   <td>v1
+   </td>
+   <td>v1
+   </td>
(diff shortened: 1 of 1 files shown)


rhuss (commented):
Looks good , but I wonder about the minimum K8s release version.

According to the [K8s versioning principles](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md), only the last 3 K8s minor versions are supported:

> Furthermore, we expect to "support" three minor releases at a time. "Support" means we expect users to be running that version in production, though we may not port fixes back before the latest minor version. For example, when v1.3 comes out, v1.0 will no longer be supported: basically, that means that the reasonable response to the question "my v1.0 cluster isn't working," is, "you should probably upgrade it, (and probably should have some time ago)". With minor releases happening approximately every three months, that means a minor release is supported for approximately nine months.

Then, K8s 1.21 is due to [April 8th](https://www.kubernetes.dev/resources/release/). So I wonder whether we shouldn't increase the minimal k8s version to 1.19 for Knative 0.22 (to be released on April 6th) ?

vaikas:
@rhuss Thanks, yeah, was actually wondering about that as well. I was curious though about the fact that we only supported .18 for one release (.21) and then dropped the support. @dprotaso I recall you have been doing some digging into the k8s version support in the past, thoughts? I have no problem changing.

n3wscott (commented):
/lgtm
/approve

dprotaso:
Our K8s min principle states
> We (the community) support the range of upstream minor Kubernetes versions that are Kubernetes community supported at the time of the cut. 

1.18 isn't EOL'd until 2021-04-30: see https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md


Also 0.23.x should be 1.19.x

dprotaso:
see: https://github.com/knative/community/issues/245#issuecomment-718316003

vaikas:
Great, thanks. Updated, @dprotaso look good now?

mattmoor (approved):
/lgtm

PR #477, opened 2021-02-19
Update toc alias in OWNER_ALIASES (adding rhuss)

rhuss:


rhuss:
@knative/steering-committee @knative/technical-oversight-committee I think as this touches a top-level file, someone from the steering committee would need to approve this PR, please.

PR #510, opened 2021-03-01
Add peribolos note

vaikas:
Add a section to the Member section that one needs to be added to the peribolos config files.


mattmoor (approved):
/lgtm

PR #512, opened 2021-03-02
Add Security Response Policy

julz:
# Changes

- :gift: Adds a Security Response policy to be followed when responding to vulnerability disclosures, based heavily on [the Envoy process](https://github.com/envoyproxy/envoy/blob/main/SECURITY.md) and Google's [OSS Vulnerability Guide](https://github.com/google/oss-vulnerability-guide/blob/main/guide.md).

This is very much a first draft open to significant modification, but I wanted us to have something solid to start discussion around.

/hold This should land after the Threat Model, Embargo List Process, and Security.md are landed (since it should be updated to link to them).

cc @evankanderson @richterdavid @AlexandraRoatis


evankanderson:
Add an OWNERS file?

security/responding.md
```diff
@@ -0,0 +1,143 @@
...
+   to relevant PRs that introduced the issue, if relevant, and any critiques of
+   the response and release process.
+1. Maintainers and Fix Team are also encouraged to send their own feedback on
+   the process to knative-dev. Honest critique is the only way we are going to
+   get good at this as a community.
+
+## References
+
+This document borrows heavily and gratefully from the [Envoy Vulnerability
+Response Process](https://github.com/envoyproxy/envoy/blob/main/SECURITY.md)
```
richterdavid, line 142: Cite the licenses these are used under?

security/responding.md
```diff
@@ -0,0 +1,143 @@
...
+   should in all cases be asked if they would like credit for identifying the
+   vulnerability in the CVE and disclosure**.
+1. The Fix Team urgently develops a fix or mitigation so that a realistic
+   timeline can be communicated to users.
+1. If the CVSS score is below 4.0, the Fix Team may - after appropriate
+   discussion and based on developer bandwidth, public holidays etc - decide to
+   slow down the release process.
+1. The fact that a vulnerability has been disclosed and a fix will be released
+   is communicated by the Fix Lead (working with the Fix Team) to the
+   knative-security-announce mailing list with an exact date and time planned
```
evankanderson, line 83: I think this needs more clarity somewhere (not this doc).

* Who is/can be on `knative-security-announce`? Is the list embargoed?
* Is this a "there is a vulnerability, and there will be a fix" message, or is this post-fix / at the same time is the fix merges?

security/responding.md
evankanderson: Since these are referenced/prior art, rather than copied, I'm not sure that license refs are needed. (In particular, it's expected that this document and those documents will drift over time.)

evankanderson (commented):
Two real comments, and some unpacking-what's-in-my-head comments just for you to read.

We'll need someone from steering to `/approve` any PRs until we have an OWNERS file under `security`.

security/responding.md
```diff
@@ -0,0 +1,143 @@
+# Vulnerability Discolure Response Policy
+
+The project follows the following process when responding to vulnerabilities
+raised via the Security Disclosure Process.
+
+## Product Security Taskforce Membership
+
+The subset of maintainers who are subscribed to the security@knative.team email
+address form the Product Security Taskforce.
```
richterdavid, line 9: why PST rather than VMT?

julz: PST is the form of words used in the envoy doc. I'm totally open to either if you feel strongly VMT is better.

security/responding.md
```diff
@@ -0,0 +1,143 @@
...
+security disclosure / email thread. This group becomes the Fix Team for the
+vulnerability.
+
+Note: Given the current size of the project it is likely that in many cases the
+Fix Team will be the entire PST, possibly with an additional engineer or
+engineers for the affected area if the PST does not have relevant expertise
+itself.
+
+If the Fix Team decides the disclosure is in fact a vulnerability which should
+be fixed, the Fix Lead will create a Github Security Advisory and provide the
```
richterdavid, line 52: Only repo admins can create a security advisory. https://docs.github.com/en/github/managing-security-vulnerabilities/about-github-security-advisories

Does this mean we make all fix leads admins? or we require PST members to be admins? Or we need to loop in an admin to create the SA? Or what?

security/responding.md
julz: Ack. The answers (I think) are (a) it's open to all (b) this is only a "prepare your engines to apply a fix on this date" message, we do not release the fix at this point since we need to inform embargo, and give people notice (to avoid a case where you have people exploiting before people notice the email and patch), but we should make sure to clarify this list and the nature of what will go to it wherever we end up posting about it.

security/responding.md
```diff
@@ -0,0 +1,143 @@
...
+## Identifying a Fix Lead
+
+The PST Fix Lead Role is responsible for initially responding to
+vulnerabilities sent to security@knative.team, and for coordinating and
+managing the response with the rest of the PST. 
+
+We use a rota system to assign the default PST Lead role. 
+All members of the PST are expected to sign up to the rota.
+Regardless of the rota, another member of the PST may volunteer to be the PST
+lead for a particular vulnerability.
```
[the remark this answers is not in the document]
julz: Willdo

security/responding.md
```diff
@@ -0,0 +1,143 @@
...
+1. Note: This step is optional and should be reserved for significant
+   vulnerabilities that warrant this treatment. In general these are
+   vulnerabilities that could lead to remotely exploitable or privilege
+   escalation issues.
+1. In the unlikely event that a member of the PST list breaks embargo, or if
+   the exploit otherwise becomes public early, the PST Lead will decide whether
+   to accelerate the schedule for public disclosure and release of the fix.
+   When in doubt push forward and go public ASAP.
+
+### Within 1-21 Days of Disclosure (Fix Release Day)
```
richterdavid, line 111: 21 days seems inconsistent with notifying the embargo list on day 14, and waiting another 3 weeks for public disclosure.

security/responding.md
```diff
@@ -0,0 +1,143 @@
...
+
+1. The Fix Lead cherry-picks the fix on to the main branch and all relevant
+   release branches and creates PRs.
+1. Maintainers should approve this PR as quickly as possible (it should already
+   have been reviewed on the private branch).
+1. If neccessary, Release jobs should be manually triggered to ensure
+   a release is available as quickly as possible.
+1. The Fix Lead completes the Github Security Advisory and requests a CVE from Github.
```
richterdavid, line 119: This seems really late in the process to get a CVE. That's documented as taking up to 3 days. I would expect that prior to the notification in line 82, so that it can be included in that communication, or concurrent with that if we don't want to wait.
https://docs.github.com/en/github/managing-security-vulnerabilities/publishing-a-security-advisory#requesting-a-cve-identification-number

security/responding.md
julz: ha! Good point. The lower bound can be achieved when the embargo step is skipped (for non-remotely-explotable/low severity vulnerabilities etc), but the upper bound is clearly longer than 21 days here. Will fix.

security/responding.md
julz: Ack, will move this around.

security/responding.md
julz: 🤔 good question.

evankanderson: I think it's reasonable to loop in a repo admin to start with once we've decided that a GHSA is needed. (Note that take-backs require going to GitHub support, so we need to be reasonably sure at this point.)

The existing repo admins will be able to see the GHSA anyway.

richterdavid: How much will they be able to see? Just the extence, or the full repo and all commits, or something else? Because the more they can see, the more we are effectively adding the repo admins to the PST for that repo. And PST membership implies a certain level of trust, which is not the same as the level of trust we put in repo admins.

evankanderson:
Per @vaikas in #516 , move these to `working-groups/security/...` to organize the top-level directories a bit more.

julz:
Rebased, moved to new working-groups/security dir, and _I think_ addressed all outstanding comments

evankanderson (commented):
/lgtm

One typo

working-groups/security/responding.md
```diff
@@ -0,0 +1,147 @@
+# Vulnerability Discolure Response Policy
```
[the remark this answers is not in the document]
julz: heh oops 😊. Turned on spellcheck and did a pass through.

evankanderson:
/lgtm
/approve

We can update the links to thread model et al as we submit those?

julz:
> We can update the links to thread model et al as we submit those?

Yep, removed the todo for now. Think this is good to go as a first pass \o/.

/unhold

PR #516, opened 2021-03-05
Add security subdirectory for security WG.

evankanderson:
# Changes

- Adds a subdirectory to organize Security WG documentation.

/kind enhancement


vaikas:
I'm curious why we're creating a top level directory for the security. Not arguing against it, just trying to understand since it's the first top level WG specific directory. Would it maybe make more sense to add it in working-groups/security instead of top level?

evankanderson:
(Excluding the `docs/` directory, which I'd argue was created in error...)

This is intended for contributor-facing rather than public documentation. I'm happy with either `working-groups` or a top-level directory, whichever you think is better. My thought on a top-level directory was to aid discoverability, but we can always move the words later.

evankanderson:
Updated

vaikas:
/lgtm
if @julz doesn't like it, we can revert :)

PR #524, opened 2021-03-09
Add initial threat model document

### LINKED ISSUE #541: Define a threat model for Knative

julz:
Fixes #541 

# Changes

- 🎁 Adds an initial (incomplete but we have to start somewhere!) threat model.

/kind documentation

### DIFF
--- a/working-groups/security/threat-model.md
+++ b/working-groups/security/threat-model.md
@@ -26,22 +26,27 @@ such as installing CRDs and managing namespaces (Control-plane-as-a-service).
 Currently, Knative assumes a single tenant per cluster (Cluster-as-a-service)
 as its security model.
 
-It is possible, with extra work, to secure Knative for use in
-Namespace-as-a-service environments, for example by setting up appropriate
-NetworkPolicy, using appropriate container isolation techniques, and filtering
-requests to shared components such as the activator. The project aims to
-eventually support secure use in Namespace-as-a-service environments out of the
-box, and is very receptive to PRs that improve support for
-Namespace-as-a-service clusters, but we do not currently view exploits in this
-environment as vulnerabilities for the purpose of our disclosure policy.
+While it is possible to secure Knative for use in non-single-tenant
+environments, for example by setting up appropriate NetworkPolicy, using
+appropriate container isolation techniques, and filtering requests to shared
+components such as the activator, the project does not currently consider
+itself to be secure for this use case. We would like eventually to support
+secure use in Namespace-as-a-service and Control-plane-as-a-service
+environments out of the box, and are very receptive to contributions that
+improve support for Namespace-as-a-service clusters, but we do not currently
+view exploits in these situations as vulnerabilities for the purpose of our
+disclosure policy.
 
 ## Trusted and untrusted workloads
 
-Knative is often used to implement platforms which permit users to submit code
-or containers, which the platform runs. In some cases the code is submitted
-only by trusted users -- for example in a private cluster -- whereas in others
-the code may be entirely untrusted (for example when knative is used to
-implement a multi-tenant FaaS-style system).
+Knative is often used to implement platforms which permit developers to submit
+code or containers, which the platform runs. In some cases only trusted
+developers can submit code (for example in a private cluster), whereas in
+others the code may be entirely untrusted (for example when knative is used to
+implement a multi-tenant FaaS-style system). In either case the resulting
+applications might be available only to trusted users (again for example in a
+private cluster), or to untrusted users (a single-tenant cluster providing
+internet-accessible applications).
 
 We differentiate between attacks on the system that require permission to
 create a Knative Service or Revision, and attacks that can be performed against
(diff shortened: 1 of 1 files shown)


working-groups/security/threat-model.md
```diff
@@ -0,0 +1,81 @@
...
+NetworkPolicy, using appropriate container isolation techniques, and filtering
+requests to shared components such as the activator. The project aims to
+eventually support secure use in Namespace-as-a-service environments out of the
+box, and is very receptive to PRs that improve support for
+Namespace-as-a-service clusters, but we do not currently view exploits in this
+environment as vulnerabilities for the purpose of our disclosure policy.
+
+## Trusted and untrusted workloads
+
+Knative is often used to implement platforms which permit users to submit code
```
evankanderson, line 40: Let's distinguish between _users_ who have access to only the http endpoints exposed outside the cluster and _developers_ who have RBAC access to update some objects on the kubernetes cluster.

working-groups/security/threat-model.md
```diff
@@ -0,0 +1,81 @@
...
+the code may be entirely untrusted (for example when knative is used to
+implement a multi-tenant FaaS-style system).
+
+We differentiate between attacks on the system that require permission to
+create a Knative Service or Revision, and attacks that can be performed against
+knative clusters or workloads even without permissions to directly deploy a
+workload. In a multi-tenant environment it is possible that all users who can
+access the environment will also have permission to deploy workloads.
+
+The most significant exploits are those that affect a cluster even without the
```
evankanderson, line 52: There's one additional class of attack where a system administrator is able to bypass cluster policy. I suggest we declare that out of scope at the moment. (This might apply in the future to certain node-local scheduling designs.)

working-groups/security/threat-model.md
```diff
@@ -0,0 +1,81 @@
...
+ability to submit workloads.
+
+## Availability, confidentiality, and integrity
+
+We consider the confidentiality of user data and integrity of user workloads
+the most significant concern. We are also concerned about Denial of Service and
+Availability attacks (particularly where exploitable without the ability to
+submit workloads), but are aware that these currently exist both within Knative
```
evankanderson, line 60: Maybe you want to say something like "these risks exist in the underlying kubernetes platform" and that we will only consider availability risks which are substantially higher than the underlying platform?

working-groups/security/threat-model.md
```diff
@@ -0,0 +1,81 @@
...
+and the underlying Kubernetes platform. For this reason availability attacks
+are currently treated as relatively lower priority.
+
+## Security of the underlying Cluster
+
+Knative relies on the security of the underlying cluster to maintain its security
+posture. This currently requires manual work that is not fully documented by
+the project (though we would eventually like to fix this).
```
evankanderson, line 68: Do _we_ want to be responsible for documenting "how to secure a k8s cluster"?

I'd rather see us link to upstream work there, and possibly an upstream a maintained security checker.

evankanderson (commented):
My big comment would be to spell out the three classes of attackers:

* users with only off-cluster access
* developers with limited RBAC access
* administrators with wide RBAC access but still subject to policy

working-groups/security/threat-model.md
julz: Sounds reasonable, Ill figure out some form of words for that

working-groups/security/threat-model.md
julz: I like that way of putting it 👍 willdo

working-groups/security/threat-model.md
julz: I would also like to see upstream work here ;). Historically I think it has been tricky (for very reasonable reasons, like backwards-compat and the generic nature of k8s) to land "secure out of the box" upstream. Hopefully this is changing.

Totally agree that to the extent we can make upstream secure such that no admin steps are needed (or refer to an existing guide to doing this) that's very much better. I somewhat suspect we will always need to recommend at least some steps, or at least point to an upstream guide / checker (I do like that idea) we expect folks to have followed.

Open to suggested ways of watering this sentence down, though..

working-groups/security/threat-model.md
```diff
@@ -0,0 +1,81 @@
...
+(Namspace-as-a-service).
+In the third, resources such as nodes and schedulers are shared, but the
+tenants, are presented with a logical cluster and can perform global operations
+such as installing CRDs and managing namespaces (Control-plane-as-a-service).
+
+Currently, Knative assumes a single tenant per cluster (Cluster-as-a-service)
+as its security model.
+
+It is possible, with extra work, to secure Knative for use in
+Namespace-as-a-service environments, for example by setting up appropriate
```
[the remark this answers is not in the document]
julz: ack, I think this paragraph was written before the CPaaS reference and then merged together. Lemme update.

working-groups/security/threat-model.md
evankanderson: There's a difference between "upstream ships secure by default" and "upstream publishes security best practices". Realistically, I'm expecting the latter for the foreseeable future.

evankanderson: Rather than saying "that is not fully documented by the project (though we would eventually like to fix this)", what about "as documented by upstream Kubernetes SIGs"?

evankanderson:
I may take a grammar pass later (e.g. there is a compound sentence where one of the clauses has two additional modifier clauses on 58-62), but let's get this in and iterate.

/lgtm
/approve

PR #529, opened 2021-03-10
Fix the Scaling WG doc

vagababov:
We never updated it after moving to the new drive

/assign @markusthoemmes @evankanderson


evankanderson (commented):
/lgtm
/approve

I thought I'd done these, but I clearly missed this one.

PR #536, opened 2021-03-18
Check in initial TOC review schedule

evankanderson:
Rather than keeping the schedule at the top of the Google doc, check it in so we can do automation (starting with https://knative.party/)

### DIFF
--- a/mechanics/TOC-REVIEW.schedule
+++ b/mechanics/TOC-REVIEW.schedule
@@ -1,0 +1,19 @@
+#@ title: ToC Working Group Update
+#@ duration: 45m
+
+# Start | Group
+2021-03-11T08:30:00-08:00 | Security WG
+# US Daylight savings
+2021-03-18T08:30:00-07:00 | User Experience WG
+2021-03-25T08:30:00-07:00 | Productivity WG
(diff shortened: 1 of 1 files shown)


evankanderson:
/assign @grantr @rhuss 

Picking two at random.

rhuss (commented):
Looks good to me. To understand the mechanics, is it that this schedule is visualized on https://knative.party ? Is there any easy way fo a WG to look up when the next date for a review is ? I think that was one major user case of the google docs table to not only see who is next, but also see when the date for my next WG update call is.

Of course I can lookup this schedule, bu this is not so easy to consume and a bit hidden.

evankanderson:
Knative.party is currently using its own copy of the file, which means that Scott Nichols needs to merge in changes. 😁 

After submitting this, I'll update https://github.com/knative-party/site/blob/main/config.yaml to point to this file, instead of one in that directory.

You can also look at the file in GitHub, it's intended to be relatively easy to read as well as for a machine to parse. If you've got suggestions to simplify, I might be able to take a swing later.

rhuss:
The format is totally fine (especially when we have the visualization on knative.party), we don't have to do it overdo it here. Thanks for taking care of the automation!

/lgtm
/approve

PR #532, opened 2021-03-12
Automate creation of OWNERS_ALIASES

evankanderson:
Automate generation of OWNERS_ALIASES. I deliberately tried to contain the `go.mod` 
blast radius to a small directory; similarly, I tried to have simple code for a 
simple feature. Happy to expand if desired/needed.

Will be followed up by a PR to http://github.com/knative-sandbox/knobots to enable 
copying `OWNERS_ALIASES` (and possibly other community files) to knative repos that 
get automation. (Another TODO: automate the addition of automation in knobots for new 
repos.)

/kind cleanup

PR #537, opened 2021-03-18
Add Ron as an actual org member

evankanderson:
See #532 -- it turns out that Ron was being removed from the org and then re-invited because he wasn't in the members list, but was on a team.


evankanderson:
Filed https://github.com/kubernetes/test-infra/issues/21434 for the lack of validation, though it looks like this might be a bit of GitHub magic going on that wasn't obvious to the Peribolos authors.

PR #552, opened 2021-03-31
Add groups based on observed usage in knative repos

evankanderson:
(PRs incoming shortly for each repo to use the groups defined here.)

Note that I did not preserve _all_ groups that existing in each OWNERS_ALIASES file. I also needed to create several groups, and split up `Serving Writers` into Working Group based groups.

My criteria for creating groups in peribolos was:
1. Associated with an existing WG (e.g. "$FOO Reviewers" for those working to approver for a WG)
2. **Multiple** directories which should have consistent ownership. There were a number of cases where a group was defined in OWNER_ALIASES and then used in only a single OWNERS file, which seems to simply make extra work. I inlined these groups in the per-repo PRs.

Note that I did _not_ try to rationalize much of the existing files, simply to migrate them to OWNERS_ALIASES

Another change which hasn't yet been done is adding in the "release leads" rotation. I expect to do that during this week, but since Markus and I have TOC approvers in most repos, I don't expect this to be a speedbump for 0.22

### DIFF
--- a/peribolos/knative.yaml
+++ b/peribolos/knative.yaml
@@ -945,7 +945,6 @@ orgs:
         - matzew
         - n3wscott
         - trshafer
-        - yu2003w
       Operations Writers:
         description: Grants write access to operations-related repositories.
         privacy: closed
--- a/peribolos/knative.yaml
+++ b/peribolos/knative.yaml
@@ -670,6 +670,32 @@ orgs:
         has_wiki: false
         homepage: https://knative.dev
     teams:
+      Autoscaling Reviewers:
+        description: Receive reviews for autoscaling directories
+        privacy: closed
+        members:
+        - taragu
+      Autoscaling Writers:
+        description: Approvers and write access for autoscaling-related repositories.
+        privacy: closed
+        repos:
+          pkg: write
+          serving: write
+        members:
+        - julz
+        - yanweiguo
+        teams:
+          Autoscaling WG Leads:
+            description: The Working Group leads for Autoscaling
+            members:
+            - vagababov
+            - markusthoemmes
+            privacy: closed
+      Client Reviewers:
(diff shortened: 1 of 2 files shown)


evankanderson:
/assign @vaikas 

(For `/approve` of root-level OWNERS_ALIASES)

peribolos/knative.yaml
```diff
@@ -848,19 +901,105 @@ orgs:
...
+        description: Receive reviews for controller utilities in pkg
+        privacy: closed
+        members:
+        - whaught
+        teams:
+          Pkg Controller Writers:
+            description: Approvers for configmap utilities in pkg
+            privacy: closed
+            members:
+            - grantr
```
dprotaso, line 989: can you add me here?

dprotaso (changes requested):

peribolos/knative.yaml
```diff
@@ -848,19 +901,105 @@ orgs:
...
+            - mdemirhan
+            - vagababov
+      Pkg Controller Reviewers:
+        description: Receive reviews for controller utilities in pkg
+        privacy: closed
+        members:
+        - whaught
+        teams:
+          Pkg Controller Writers:
+            description: Approvers for configmap utilities in pkg
```
dprotaso, line 986: typo 
```suggestion
            description: Approvers for controller utilities in pkg
```

peribolos/knative.yaml
evankanderson: Done.

peribolos/knative.yaml
evankanderson: Fixed.

vaikas (approved):
/lgtm

/hold
just adding a hold in case others want to chime in. Remove as y'all see fit.

PR #561, opened 2021-04-09
Toc2021 election Elekto changes

jberkus:
PR to add Elekto-specific files and data for the TOC election.

This will need to be in place by the 13th.  We'll also need to add a webhook to this repository.

@bsnchan @geekygirldawn


jberkus:
Feh, this dragged in a bunch of garbage with it.  Lemme rebase and start over.

PR #563, opened 2021-04-12
Toc2021 voters list

jberkus:
Add voters list for 2021 TOC election.

Voter data from TestStats.  List pulled by @geekygirldawn.

Pushing this list also tests out Elekto integration.

@bsnchan


jberkus:
Git history still broken, gonna re-create.

PR #575, opened 2021-04-19
Regenerate OWNERS_ALIASES

evankanderson:
I'm still working on knobots automation, so regenerate this by hand today.


evankanderson:
/assign @bsnchan @pmorie @thisisnotapril 

(Since steering approves the top-level OWNERS_ALIASES)

PR #580, opened 2021-04-22
changing times of Docs + UX to reflect actual meeting times

omerbensaadon:


evankanderson (commented):
Just want to check that 1 hour/week is the expected meeting time for DUX?

/lgtm
/approve

csantanapr:
@evankanderson I think 1 hour a week is ok, do you have a concern? or meeting twice 2 30minutes better?
I prefer to meet once.

PR #585, opened 2021-04-23
csantanapr TOC candidacy submission

csantanapr:
csantanapr TOC candidacy submission

### DIFF
--- a/elections/2021-TOC/candidate-csantanapr.md
+++ b/elections/2021-TOC/candidate-csantanapr.md
@@ -1,0 +1,17 @@
+-------------------------------------------------------------
+name: Carlos Santana
+ID: csantanapr
+info:
+  - affiliation: IBM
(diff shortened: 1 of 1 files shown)


csantanapr:
I was made aware I don't qualify to run for TOC.

rhuss:
@csantanapr please note that it still makes sense to run for TOC, even when constraints would prevent you to be elected right now. However as it happened in the past, people might move on during the 2 year period, so that a backfill is picked from this election (in the next year), so the constraints are evaluated afresh.

rhuss:
My comment applies to the constraint that only 2 seats being held by employees from the same organization (or conglomerate, in the case of companies owning each other). Not to any other constraints as described in that apply to the election itself https://github.com/knative/community/blob/main/mechanics/TOC.md#candidate-eligibility (e.g. being 3 month an approver at least).

PR #586, opened 2021-04-26
Add OWNERS_ALIASES for knative-sandbox

evankanderson:
Start for: https://github.com/knative/test-infra/issues/2751

We'll use the `update-community` action in https://github.com/knative-sandbox/knobots to copy this to all the repos in knative-sandbox.


hack/update-codegen.sh
```diff
@@ -25,6 +25,7 @@ function make_aliases() {
   ( cd mechanics/tools/gen-aliases; go build . )
   # Auto-generate OWNERS_ALIASES from peribolos configs
   ./mechanics/tools/gen-aliases/gen-aliases
```
rhuss, line 27: would it make sense to add explicite arguments here, too (like in the line below), so that it's more clear which files are worked on ? (I suspect it's "knativeperibolos/knative.yaml knative-OWNERS_ALIASES" ?)

peribolos/knative-sandbox.yaml
```diff
@@ -155,7 +155,6 @@ orgs:
     - matzew
     - mayaspivak
     - maximilien
-    - mdemirhan
```
rhuss, line 158: Is this by accident or intentional? (as it does not seem to be part of the original intent of the PR)

evankanderson: It turns out that Mustafa and Tomas haven't accepted invites to the knative-sandbox org, so aren't valid members in an OWNERS_ALIASES file, which Prow checks.

hack/update-codegen.sh
evankanderson: Done.

peribolos/knative-sandbox.yaml
evankanderson: (Also, Mustafa has switched teams/companies and headed back to Microsoft; Tomas accepted a role in a different part of Google.)

peribolos/knative-sandbox.yaml
```diff
@@ -639,11 +638,11 @@ orgs:
           Steering Committee:
             description: ""
             maintainers:
-            - isdal
             - pmorie
             - bsnchan
             - mbehrendt
             - thisisnotapril
+            - vaikas
```
vaikas, line 645: Should SC also have a 'fallback-notification-blocker' in their?

vaikas (approved):
/lgtm
/hold

Thanks! If the fallback is not necessary, feel free to unhold at will.

PR #592, opened 2021-04-28
Add two voters per approved exception requests.

jberkus:
@bsnchan, please approve.  Thanks!


bsnchan:
For context: The steering committee reviewed these two exceptions and agreed that these two candidates should be able to vote.

/lgtm
/approve
/ok-to-test

PR #594, opened 2021-04-29
Add section about updating docs

vaikas:
# Changes

- :broom: When removing repos, added a note saying that we should also update documentation so there's no dangling pointers around.
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


evankanderson (commented):
/lgtm
/approve

PR #598, opened 2021-05-05
Remove emeritus members from teams

evankanderson:
mdemirhan@ is no longer in the Knative org; removing him from teams to avoid OWNERS failures.


vaikas (approved):
/lgtm

PR #615, opened 2021-05-19
Update knative.yaml

bbtong:
Hello! I am from VMware, requesting to add `bbtong` to members


bbtong:
need to sign CLA First

PR #620, opened 2021-05-20
Revisit rules around what happens when a TOC member resigns

julz:
The current situation is silly :). Because, in the event of a TOC member resigning or leaving, the next in line from the previous election is used, folks are encouraged to run in elections for seats that - due to vendor limits - aren’t available to them, just for purposes of vacancies that may arise during the year. This year that has become even, frankly, sillier since an extra resignation has meant two such elections will be held in quick succession.

The current mechanic likely made sense when there were more independent vendors in the project, since in that case the situation was substantially less likely to arise. Now it seems like we'll quite commonly (possibly every election, unless the balance of contributions substantially changes, given the make-up of upcoming vacancies means all IBM/RH vacancies and then all VMware vacancies happen on alternate years) have many candidates that are ineligible, unless we change this rule (and/or the handling of vendor limits).

We can remove the need for people to run for seats that aren’t available by simply holding a special election (rather than using the previous results) in the event a seat does become available. This will avoid encouraging us to waste folk’s time and (frankly) emotional energy when a seat isn’t available, and would in any event (IMO, anyway) be a better and fairer way to fill any vacancies, especially if they arise later in the year.

### DIFF
--- a/mechanics/TOC.md
+++ b/mechanics/TOC.md
@@ -32,10 +32,6 @@ with at least 3 months tenure are eligible to stand for election. Candidates may
 self-nominate or be nominated by another eligible member. The approximate time
 commitment of a TOC member is around 8 hours per week.
 
-Please note that vendor or organization limits do not restrict candidate eligibility.
-We encourage all interested candidates in running as there may be occurrences where a TOC member
-steps down and is back-filled by the candidate with the next most votes from the previous election.
-
 # Voter Eligibility
 
 Anyone who has at least 50 contributions in the last 12 months is eligible to
@@ -74,16 +70,10 @@ committee elections.
 
 ## Vacancies
 
-In the event of a resignation or other loss of an elected TOC member, the
-candidate with the next most votes from the previous election will be offered
-the seat. This process will continue until the seat is filled.
-
-In case this fails to fill the seat, a special election for that position will
-be held as soon as possible. Eligible voters from the most recent election will
-vote in the special election (ie: eligibility will not be redetermined at the
-time of the special election). Any replacement TOC member will serve out the
-remainder of the term for the person they are replacing, regardless of the
-length of that remainder.
+In the event of a resignation or other loss of an elected TOC member, a special
+election for that position will be held as soon as possible. Any replacement TOC
+member will serve out the remainder of the term for the person they are replacing,
+regardless of the length of that remainder.
 
 ---
 
--- a/mechanics/TOC.md
+++ b/mechanics/TOC.md
@@ -70,23 +70,23 @@ committee elections.
 
 ## Vacancies
 
-In the event of a resignation or other loss of an elected TOC member with more
-than three months left until the next regular annual election, the candidate
-with the next most votes from the previous annual election will be offered the
-seat. If the seat cannot be filled from the previous annual election, the
-candidate with the next most votes from the most recent special election will
-be offered the seat. This process will continue until the seat is filled.
+In the event of a resignation or other loss of an elected TOC member with less
+than three months left until the next regular annual election, the Steering
+Committee shall appoint a qualified contributor to fill that TOC seat until
+that election.
+
+Otherwise the candidate with the next most votes from the previous annual
+election will be offered the seat. If the seat cannot be filled from the
+previous annual election, the candidate with the next most votes from the most
+recent special election will be offered the seat. This process will continue
+until the seat is filled.
 
 In case this fails to fill the seat, a special election for that position will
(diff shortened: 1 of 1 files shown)


geekygirldawn:
My biggest concern with this is that it could actually result in a lot of elections, which are incredibly time consuming for the steering committee, the community, and your election officers. Looking at the history of the TOC, it looks like there have been 3 resignations since February, and I don't think any of us want that many TOC elections.

julz:
> My biggest concern with this is that it could actually result in a lot of elections, which are incredibly time consuming for the steering committee, the community, and your election officers.

This is fair, I for one would like not like more elections! :D I'd just point out though the under the _current_ rules we've already ended up with two elections in a row, which is pretty disruptive if everyone has to run in each one just due to the fallback rule when they're not eligible.

An alternative solution would be simply to add the word "annual" so the rule would read "In the event of a resignation or other loss of an elected TOC member, the candidate with the next most votes from the previous **annual** election will be offered the seat.". This would not result in extra elections, but would at least avoid asking candidates who are not eligible to fill the seat to stand again when a special election needs to be called to fill a seat.

julz:
(FWIW I think this has only become a big issue given how frequently I think this situation is now likely to arise, given the balance of contributions and the fact that the cadence of vacancies is now really unfortunate, i.e. IBM/RH every 2 years, VMW every 2 years. If we do something about #621 I'd probably feel differently about this issue).

mechanics/TOC.md
```diff
@@ -74,16 +70,10 @@ committee elections.
...
-In case this fails to fill the seat, a special election for that position will
-be held as soon as possible. Eligible voters from the most recent election will
-vote in the special election (ie: eligibility will not be redetermined at the
-time of the special election). Any replacement TOC member will serve out the
-remainder of the term for the person they are replacing, regardless of the
-length of that remainder.
+In the event of a resignation or other loss of an elected TOC member, a special
+election for that position will be held as soon as possible. Any replacement TOC
+member will serve out the remainder of the term for the person they are replacing,
+regardless of the length of that remainder.
```
geekygirldawn, line 76: This gets me thinking about another alternative. Something like this might be a good middle ground? Changes in _italics_:

> In the event of a resignation or other loss of an elected TOC member, the candidate with the next most votes from the previous _annual_ election will be offered the seat. _If the seat cannot be filled from the  previous annual election, the candidate with the next most votes from the most recent special election will be offered the seat._ This process will continue until the seat is filled. In case this fails to fill the seat, a special election for that position will be held as soon as possible. Eligible voters from the most recent election will vote in the special election (ie: eligibility will not be redetermined at the time of the special election). Any replacement TOC member will serve out the remainder of the term for the person they are replacing, regardless of the length of that remainder.

julz: yeah, that would look good to me 👍

jberkus:
Suggestion as an additional para, and a way to avoid excess elections that's been used in other projects:

"In the event that a TOC seat becomes available with three months or less left until the next regular annual election, the Steering Committee shall appoint a qualified contributor to fill that TOC seat until that election."

This avoids the circumstance where someone resigns in early March and we have to hold a special election or leave the seat unoccupied.

mechanics/TOC.md
vaikas: @julz did you want to update this section and add the section here that @jberkus suggested:
```
"In the event that a TOC seat becomes available with three months or less left until the next regular annual election, the Steering Committee shall appoint a qualified contributor to fill that TOC seat until that election."
```

julz: 👍 Done. Updated with Dawn and Josh's suggestions

mechanics/TOC.md
```diff
@@ -85,6 +84,10 @@ time of the special election). Any replacement TOC member will serve out the
 remainder of the term for the person they are replacing, regardless of the
 length of that remainder.
 
+In the event that a TOC seat becomes available with three months or less left
```
vaikas, line 87: bikeshedding: I wonder if we should move this up front (to L73). Then I think we don't have to have the L73 state the more than three months. That would then be else clause?
right now it's kind of like this:
if > 3 months
    - annual election order
 if that didn't work out
    - special election
 if <= 3 months
    - appoint
    
   It might be clearer to just say:
   if <= 3 months
     - appoint
    
    else L73-L85 (minus the mentioning of 3 months, since it was already taken care of above?)

julz: ack, updated, ptal

vaikas:
/lgtm
I'll wait to approve to make sure other folks get a chance to look. Thanks for doing this!

PR #618, opened 2021-05-19
[WIP] Creating initial structure and content for the 2021 special election

geekygirldawn:
This should not be merged until the following 2 conditions are met:

1. @jberkus to confirm that this meets the Elekto requirements and that it otherwise looks good.
2. Steering to confirm that the instructions and deadlines in the readme meet with your approval.

cc: @bsnchan @pmorie @thisisnotapril @mbehrendt @vaikas

PR #625, opened 2021-05-20
Remove company limits for TOC

### LINKED ISSUE #621: PROCESS CHANGE: Revisit representation limits on TOC
What the subject says really :).

I'm super super super biased since this recently affected me, but I really do think the current situation was designed for a project with a different balance of contributions and needs to be revisited. The recent elections were painful and the same situation seems likely to repeat. At minimum I think now that we have separate steering and trademark committees, with their own representation limits (neither of which existed when the TOC was formed) we should add more flexibility to the rules around conglomerates (or representation limits in general), at least for the TOC. 

Other projects, for example Tekton, use the following form of words to evaluate independence of subsidiaries, which would at least allow more flexibility "[In the event of a question of company membership (for example evaluating independence of corporate subsidiaries) a majority of all non-involved Governance Board members will decide](https://github.com/tektoncd/community/blob/main/governance.md#maximum-representation)". 

Kubernetes also uses a substantially similar rule, "[in the event of a question of company membership (for example evaluating independence of corporate subsidiaries) a majority of all non-involved Steering Committee members will decide"](https://github.com/kubernetes/steering/blob/master/elections.md#maximal-representation).

Especially now that we have separate steering and trademark committees it at least seems reasonable to add flexibility around conglomerates at TOC level, matching both Kubernetes and Tekton in this area, rather than a blanket rule. (Tho TBH I'd personally also be in favour of just removing representation limits at the TOC level entirely, given we have steering and trademark now).

julz:
Fixes #621 

As discussed at steering meeting 20/5/21.

### DIFF
--- a/mechanics/TOC.md
+++ b/mechanics/TOC.md
@@ -17,8 +17,7 @@ aliases:
 
 # Composition
 
-The TOC will have five seats with a 2 year term, with a maximum of 2 seats being
-held by employees from the same organization (or conglomerate, in the case of companies owning each other).
+The TOC will have five seats with a 2 year term.
 
 There will be an annual election to determine the composition of the TOC for the
 following year. Three seats will be up for election in one year and two will be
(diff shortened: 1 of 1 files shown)


julz:
cc: @bsnchan @pmorie @thisisnotapril @mbehrendt @vaikas

julz:
We will also (iiuc) need to update the [WIP PR for the special election](https://github.com/knative/community/pull/618), since my understanding is this will change the situation there.

cc @jberkus @geekygirldawn fyi

vaikas (commented):
/lgtm
Thanks!

vaikas (approved):

PR #632, opened 2021-05-26
Adding eric-sap as eventing-kafka approver

travis-minke-sap:
I would like to add `eric-sap` as an eventing-kafka approver.  @matzew has also indicated his agreement with this ; )

I'm not up-to-speed as to whether this is the correct config and whether anything else might be required - please advise if so - thanks!

### DIFF
--- a/peribolos/knative-sandbox.yaml
+++ b/peribolos/knative-sandbox.yaml
@@ -87,6 +87,7 @@ orgs:
     - eobrain
     - ericbottard
     - ericlem
+    - eric-sap
     - fabolive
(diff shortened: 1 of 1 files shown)


peribolos/knative-sandbox.yaml
```diff
@@ -860,6 +861,7 @@ orgs:
         - aliok
         - davyodom
         - devguyio
+        - eric-sap
```
lionelvillard, line 864: we don't use CODEOWNERS anymore. Anyway, no harm to add it here

PR #640, opened 2021-06-01
Updated Elekto results page with the final TOC election results

geekygirldawn:
Updated Elekto results page with the final TOC election results, since we are not having a runoff.

@jberkus Can you have a quick look to make sure that this looks good to you and doesn't break anything in Elekto before we have someone from steering approve it?

/hold


evankanderson (commented):
/lgtm
/approve

We should have put you two in `/elections/2021-TOC/OWNERS` to simplify your workflow. Thanks for all your work!

jberkus:
Looks good, please merge.

PR #641, opened 2021-06-03
Add bbtong to the members list

bbtong:
Add `bbtong` to the Knative members list, as part of VMware CNR Connectivity to eventually submit to Knative feature work.

I signed the VMware<>Google internal CCLA, as was advised to do that, but (VMware someone?) let me know if I still have to sign Google's CLA beyond it.


bbtong:
@googlebot I signed it!

PR #653, opened 2021-06-15
[repo-request] knative-sandbox/monitoring

dprotaso:
The purpose is to revive aspects of the monitoring bundle - see https://github.com/knative/community/issues/654

cc @upodroid

### DIFF
--- a/peribolos/knative-sandbox.yaml
+++ b/peribolos/knative-sandbox.yaml
@@ -454,6 +454,12 @@ orgs:
         description: A benchmarking tool to evaluate Knative performance
         has_projects: true
         has_wiki: false
+      monitoring:
+        allow_merge_commit: false
+        allow_rebase_merge: false
+        description: Observability support for Knative
+        has_projects: false
+        has_wiki: false
       net-certmanager:
         allow_merge_commit: false
         allow_rebase_merge: false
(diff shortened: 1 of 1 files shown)


dprotaso:
cc @skonto, @itsmurugappan, @yuzisun should I add you as reviewers/writers to this repo?

upodroid:
/retest
I accepted the invites

peribolos/knative-sandbox.yaml
```diff
@@ -701,7 +708,11 @@ orgs:
       Operations Writers:
         description: Grants write access to operations-related repositories.
         privacy: closed
-        repos: {}
+        repos:
+          monitoring: write
+        members:
+        - dprotaso
+        - upodroid
```
evankanderson, line 715: My $0.02 is that for a single repo, we don't need to create a group. This list should be actual "approvers" in the Operations WG.

I'd add `upodroid` to OWNERS in the resulting repo rather than adding them to (possibly multiple) experiments here in the future.

dprotaso: Actually thinking a bit - should we aim to move one off users out of OWNER files and into GitHub teams just for the sake of making removal/cleanup easier?

evankanderson (commented):
/lgtm
/approve

PR #677, opened 2021-07-07
Remove test and to-import repo from peribolos

evankanderson:
Also rerun ./hack/update-autogen.sh, until Evan gets around to automating it.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -102,6 +102,7 @@ aliases:
   - julz
   - lionelvillard
   - markusthoemmes
+  - matzew
   - n3wscott
   - nak3
   - navidshaikh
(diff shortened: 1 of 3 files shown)


rhuss (commented):
/lgtm
/approve

rhuss:
@knative/steering-committee the PR touches a top-level file as well.

Also, there are some other changes in the peribolos file, so maybe we need an other run to regenerate the top-level aliases ?

Automation would be cool as this would also free us ping steering everytime a change in the peribolos config happens.

evankanderson:
Yeah, automation is on my list, probably after https://github.com/knative-sandbox/knobots/issues/94

rhuss:
@evankanderson could you please re-run the script and update the PR ? That has been two updates earlier today to the peribolos config, so would be cool to get that in in one go.

evankanderson (commented):
Updated! So long, Slinky, and thanks for all the 🐟 !

PR #684, opened 2021-07-13
Remove stray occurrences of kn-plugin-func

markusthoemmes:
As per title, these have been missed and trip the nightly runs.


rhuss (commented):
/lgtm
/approve

thanks !

Issue #648, opened 2021-06-14
PROCESS CHANGE: Add "Are you using Knative?" issue to Community, link to it, maintain an ADOPTERS.MD in community 


omerbensaadon:
Proposal: Create a "Are you using Knative?" issue in Community similar to https://github.com/dapr/dapr/issues/3169, pin it. 

Create a similar issue in each of our main repos (serving, eventing, docs), pin it, links to this issue in Community. 

See the proposed text: 
```
# Are you using Knative?

If you are using Knative, we would love to know!

## Why?
Tracking usage in OSS is hard, if you leave a comment on this issue we can better understand who our users are and deliver a better experience for YOU!

- We'd like to give your organization the opportunity to collaborate more closely with contributors (if you want to)
- We want to help promote your organization 
- We want to show off all the awesome stuff our users are doing 

## How you can help
Submit a comment in this issue to include the following information, which would then also get included into the [ADOPTERS.md]($//TODO LINK) file and (if applicable) on the Knative homepage: 

Organization/Company: 
Website: 
Country: 
Contact: (email or Twitter at least one please!)
Usage scenario: e.g. We use Knative for XYZ
Status: Development?/Production?
```

* Expected benefits. Who gains the benefits? Why will they benefit?
> We get a better understanding of who our users are, more logos for the website, more reference implementations etc. etc. 
* Expected costs. Who bears the costs? How heavy are they?
> Next to nothing 
* Timeframe for implementation / rollout.
> As soon as it is approved
* Are you willing to drive the process, or is this a request for help?
> Yes, I need someone to pin the issue in each Repo though (I believe)

omerbensaadon:
I won't be able to attend the next 2 Steering Meetings :(

xtreme-sameer-vohra:
Related - https://github.com/knative/docs/issues/3246

bsnchan:
cc @knative/steering-committee 

I love this idea! I can bring it up at the next steering meeting if you can't make it :)

bsnchan:
Steering loves this idea :) let's do it

/assign @bsnchan

bsnchan:
Main issue created here (https://github.com/knative/community/issues/688), will update the issue once I have the ADOPTERS.MD file merged in.

bsnchan:
Issue has been created -- I pinned it at the top of of knative/community. @omerbensaadon - were you thinking to create an issue in serving/eventing/docs as well? I'm worried that the comments may end up in different places if we were to do that. What do you think?

[...]

PR #689, opened 2021-07-14
Creates ADOPTERS.MD file

bsnchan:
# Changes

Create ADOPTERS.MD file as part of https://github.com/knative/community/issues/648

### DIFF
--- a/ADOPTERS.MD
+++ b/ADOPTERS.MD
@@ -1,0 +1,11 @@
+# Knative Adopters
+
+This page contains a list of organizations who are using Knative in production or in stages of development.
+
+If you would like to be included in this table, please submit a comment to [this issue](https://github.com/knative/community/issues/688) and your information will be added.
+
+## Adopters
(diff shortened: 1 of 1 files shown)


ADOPTERS.MD
```diff
@@ -0,0 +1,11 @@
+# Knative Adopters
+
+This page contains a list of organizations who are using Knative in production or in stages of development.
+
+If you would like to be included in this table, please submit a comment to [this issue](https://github.com/knative/community/issues/688) and your information will be added.
+
+## Adopters
+
+| Organization | Status | More Information
```
vaikas, line 9: Just a nit (seems to display correctly in the rich view), but shouldn't the '|' before More Information be indented one more over and shouldn't there be a trailing '|'?

bsnchan: I don't think it makes a difference, but let me update so it's not confusing

Issue #626, opened 2021-05-21
Create a process for Emeritus *


vaikas:
As mentioned in #622 (and I think in few other places (or at least in my head it was brought up before)) the topic of adding an Emeritus role to give people that no longer have the role but have done it in the past the recognition that they deserve.

At the very least, I think WG leads, TOC, Steering, KTC members should have that, this is just from the top of my head and maybe incomplete.

Maybe we could add a table below (for example TOC):
https://github.com/knative/community/blob/main/TECH-OVERSIGHT-COMMITTEE.md#committee-members

and list the Emeritus there.

rhuss:
+1 from my side, we could it individually on every level (TOC, SC, TMC, WGs) .... Should we add the years of activity, too ? (but that's probably overkill)

vaikas:
Aaand finally for WG leads 'process', we had: #701 

So, if WG leads could go and update their sections as appropriate that would be swell. Please update your WG section, and check the checkboxes once they've been done so we can close this once all the boxes have been checked. If your WG has no Emeritus members, please still check the box so I know when all WG have looked at this.

[ ] [API Core](https://github.com/knative/community/blob/main/working-groups/WORKING-GROUPS.md#api-core)
[ ] [Client](https://github.com/knative/community/blob/main/working-groups/WORKING-GROUPS.md#client)
[ ] [DUX](https://github.com/knative/community/blob/main/working-groups/WORKING-GROUPS.md#documentation--user-experience)
[ ] [Event Delivery](https://github.com/knative/community/blob/main/working-groups/WORKING-GROUPS.md#event-delivery)
[ ] [Eventing Sources](https://github.com/knative/community/blob/main/working-groups/WORKING-GROUPS.md#eventing-sources) 
[ ] [Networking](https://github.com/knative/community/blob/main/working-groups/WORKING-GROUPS.md#networking)
[ ] [Operations](https://github.com/knative/community/blob/main/working-groups/WORKING-GROUPS.md#operations)
[ ] [Productivity](https://github.com/knative/community/blob/main/working-groups/WORKING-GROUPS.md#productivity)
[ ] [Scaling](https://github.com/knative/community/blob/main/working-groups/WORKING-GROUPS.md#scaling)
[ ] [Security](https://github.com/knative/community/blob/main/working-groups/WORKING-GROUPS.md#security)

PR #692, opened 2021-07-15
Add proposed format for emeritus for TOC

vaikas:
As discussed in #626 wanted to put a possible format for TOC for discussion. If this is aok, I'll make these changes elsewhere as well.

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


julz (approved):
Nice. Love this ❤️ 

/lgtm
/hold just to give people time to bike shed if anyone wants to, but @vaikas feel free to unhold whenever you like as far as I'm concerned

vaikas:
The one reason I left out the company affiliation is because for some folks it has changed since they left, changed while the folks were serving, etc. and would've complicated things for no benefit that I could see.

julz:
> The one reason I left out the company affiliation is because for some folks it has changed since they left, changed while the folks were serving, etc. and would've complicated things for no benefit that I could see.

I guessed that was why, fwiw, and it makes sense to me

bsnchan:
/approve

Thanks for doing this <3 !

csantanapr:
❤️ this idea @vaikas +1

vaikas:
Ok, I think it's been long enough for folks to comment. We can always change things later as we see fit.
/unhold

PR #716, opened 2021-07-29
Drop last occurrence of peribolos test repo

markusthoemmes:
As per title. This currently fails the peribolos jobs. See https://prow.knative.dev/view/gs/knative-prow/logs/post-knative-sandbox-peribolos/1420676713062338560


rhuss (commented):
/lgtm
/approve

oops, didn't know that this was left over.

PR #722, opened 2021-08-02
Remove @vagababov as Scaling WG lead

vagababov:
After changing teams and furthermore after 2.5 months of parental leave,
I just don't feel like I can contribute at the level this position
requires.

It's been a please and an honor serving in this position.

I will be still be around, still review code, and what not, but it's
120% project, than a full time job for me now.

Sorry, Markus, than I am increasing your workload, though probably temporarily :-)

Thanks, yours truly.


julz:
Just want to say I learnt an absolute ton working with you @vagababov (I know you're not gone and will still be reviewing etc when you can, but I still wanna say it ❤️)

vaikas:
Thanks for everything @vagababov ❤️ 
/lgtm
/approve

Issue #731, opened 2021-08-12
PROCESS CHANGE: Introduce end-user seats in steering committee


bsnchan:
## Background
When we transitioned from the bootstrap steering committee to an elected steering committee last year (in 2020), we explored the idea of creating an end-user seat but couldn't quite determine what mechanics we should follow. We [documented](https://github.com/knative/community/blob/main/STEERING-COMMITTEE.md#transition) that we should explore this process again for the 2021 steering elections.

## Proposed Change
* Keep 5 total seats on steering, 4 contributor seats + 1 end-user seat.
* For 2021 steering elections, we have the 3 remaining bootstrap seats up for election. Have 2 of these seats be contributor seats and 1 of these seats be the end-user seat.
* Proposal for election process options:
  * For 2021: Have steering appoint the end-user representative who will stay on the committee for a 1-year term. If the representative leaves their company, stops being an end-user, steps down, steering can reappoint as necessary (thanks @julz for the idea!)
  * For 2022: Create an "end user registry" where users share their user stories. This will give us a pool of "end users" which the community could then vote on (thanks @jberkus for the idea!)

## Benefits
* Lower the bar for end-users to join the community
* Increase representation in the steering committee

## Cost
* We would be losing 1 contributor seat (out of 5 seats)

## Timeframe for implementation/rollout
* We can adopt this process and start evangelizing/sharing before the steering committee elections in late Oct/early Nov
* I don't believe it will require additional work for the election officers

## Are you willing to drive the process, or is this a request for help?
* Yes

vaikas:
If we do not get someone to run, perhaps fall back to contributors?

geekygirldawn:
I'm curious about how we implement this from a voting perspective. Will other end users be able to vote on the end user seat? If so, will end users also be able to vote on the contributor seat? Or do we leave the voting criteria as-is?

@jberkus from an Elekto perspective, would we need to run this as 2 separate elections - one for the end user seat and one for the contributor seats? Or do we run it as one election with people ranking both types of candidates together and select the top 2 contributors and the top end user? This decision impacts @vaikas's PR.

vaikas:
I'm curious about this:

```
The end-user seat would be the "tie-breaker" if there were any decisions that needed tie-breaking
```

[...]

PR #734, opened 2021-08-18
2021 SC elections mechanics

vaikas:
:gift: 2021 SC Elections. Still work in progress.

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


vaikas:
@geekygirldawn @jberkus

elections/2021/SC/README.md
```diff
@@ -0,0 +1,163 @@
...
+The role of this election is to fill out the three (3) seats due for election
+this year on the [Knative Steering Committee]. Each elected member will serve a
+two (2) year term.
+
+## Background
+
+This election will shape the future of Knative as a community and project. The
+Knative Steering Committee (SC) is responsible for the general health of the
+Knative community. Seats up for the election are the remaining bootstrap seats
+and in addition to contributor seats, we're introducing one (1) end-user seat to
```
bsnchan, line 21: Do we want to make sure https://github.com/knative/community/issues/731 gets merged first?

bsnchan: Also, based on #731 - the proposed change is that we have the 3 remaining bootstrap seats up for election. This change would mean we would have 2 of these seats be contributor seats and 1 of these seats be the end-user seat.

vaikas: This is mosdef wip still and gated on that, so yupp

geekygirldawn:
@jberkus - Do we need to rename the dir to 2021-SC? or has Elekto been updated to handle other directory structures?

elections/2021/SC/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+The list of eligible voters is stored (TODO: vaikas add this)[in the community repository](./voters.yaml), but it is easier
+to check if you are eligible by logging into [Elekto]. See Voting Process below for more details.
+
+## Schedule
+
+| Date         | Event                    |
+| ------------ | ------------------------ |
+| October 25   | Announcement of Election, call for nominations |
+| November 8   | All candidate nominations and voting exception forms due by 0000 UTC (5pm Pacific) |
```
geekygirldawn, line 39: Since the exception forms are handled in Elekto, we should change this to:
| November 8   | All candidate nominations due by 0000 UTC (5pm Pacific)

and add this one:
| November 19 | Voter exception requests due by 0000 UTC (5pm Pacific)

I pulled the date in a bit so that we could handle the exceptions in the week before the US Thanksgiving holiday.

vaikas:
> @jberkus - Do we need to rename the dir to 2021-SC? or has Elekto been updated to handle other directory structures?

Oh snap, I didn't realize it was Elekto requirement. I just noticed in 2020 we had 2020/{SC, TOC} dirs. Happy to rename, not a big deal 👍

elections/2021/SC/README.md
vaikas: Done 👍

jberkus:
@vaikas actually, I'm testing supporting arbitrarily deep nested directories right now.  should be working by the time you need it.

bsnchan:
Hi @jberkus, could you accept your invitation to the knative github org please? Thanks in advance!

vaikas:
Do folks have a preference for directory structure? I made the change so it's not 2021-SC but I can change it back to what it was. I personally think it might be nicer to have the 2021/SC but maybe we'll just change to that in 2022 since there's already 2021-TOC so doesn't make much of a difference.

vaikas:
Since I didn't hear any preferences for the directory structure, and given how the TOC was done, I just kept it consistent with that for now. I updated based on #758 so this should be good to go for a look.

vaikas:
@geekygirldawn @jberkus @pmorie @mbehrendt @thisisnotapril 
Would you take a look please? Should have everything in there that we discussed.

geekygirldawn:
I think both @jberkus and I are back after having some time off, so we can start moving on this again. 

This looks good to me. I would like for @jberkus to have a quick look to make sure we have any Elekto requirements taken care of.

PR #735, opened 2021-08-18
add jberkus as member

vaikas:
Need to add jberkus as a member as per #734 so that @jberkus can be added as Owner for the SC elections directory.

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/peribolos/knative.yaml
+++ b/peribolos/knative.yaml
@@ -231,6 +231,7 @@ orgs:
     - janmedrek
     - jasonkeene
     - jasonpolites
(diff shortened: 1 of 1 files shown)


vaikas:
/hold
Hold to make sure @jberkus is ok with this so this doesn't get accidentally merged.

jberkus:
oh!  sorry, didn't realize I needed to do anything.

/lgtm

PR #736, opened 2021-08-18
v0.26 release leads 

lionelvillard:
# Changes

- Set @ZhiminXiang and @evankanderson as v0.26 release leads
-
-

 

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -69,7 +69,6 @@ aliases:
   - pierDipi
   - vaikas
   knative-admin:
-  - RichieEscarez
   - ZhiminXiang
   - bsnchan
   - dprotaso
@@ -112,7 +111,6 @@ aliases:
   - vagababov
   - vaikas
   knative-release-leads:
-  - RichieEscarez
   - ZhiminXiang
   - evankanderson
   knative-robots:
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -70,6 +70,7 @@ aliases:
   - vaikas
   knative-admin:
   - RichieEscarez
+  - ZhiminXiang
   - bsnchan
   - dprotaso
   - duglin
@@ -79,7 +80,6 @@ aliases:
   - knative-prow-robot
   - knative-prow-updater-robot
   - knative-test-reporter-robot
-  - lionelvillard
   - markusthoemmes
   - mbehrendt
   - pmorie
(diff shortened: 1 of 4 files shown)


evankanderson (commented):
Can you (manually) run `scripts/update-gen.sh` (I think) to update OWNERS_ALIASES as well?

lionelvillard:
done! (ran `hack/update-codegen.sh`)

OWNERS_ALIASES
```diff
@@ -114,8 +113,8 @@ aliases:
   - vaikas
   knative-release-leads:
   - RichieEscarez
-  - lionelvillard
-  - yanweiguo
+  - ZhiminXiang
```
vaikas, line 116: Just curious on why / if we still need @RichieEscarez here as a release lead. My understanding was that it was earlier required for Hugo related tasks, but release leads have been able to do this for a few releases now. See below also.

peribolos/knative.yaml
```diff
@@ -877,8 +877,8 @@ orgs:
           Knative Release Leads:
             description: Active member of the Knative Release Leads rotation.
             maintainers:
-            - yanweiguo
-            - lionelvillard
+            - ZhiminXiang
+            - evankanderson
```
vaikas, line 881: I think the next issue has now been done, so I believe we can remove that entry now and solely rely on the Knative Release Leads and no special docs permissions are required.

lionelvillard: not sure I understand. Is there anything to do other than removing Richie (which I did)?

vaikas: Nope, that was all I thought we could now safely do. I just jotted it first down in the first file, then pointed to place where (which you did in this commit) the change needed to be made.
Thanks and sorry for the confusing verbiage :)

OWNERS_ALIASES
```diff
@@ -69,7 +69,7 @@ aliases:
   - pierDipi
   - vaikas
   knative-admin:
-  - RichieEscarez
+  - ZhiminXiang
```
vaikas, line 72: I don't actually think we should be making any changes as part of this PR (for the release) to the knative-admin. Only the knative-release-leads should change.
Revisiting the knative-admin roles (I think) should be done as a separate PR.

lionelvillard: hmm this is automatically generated.... maybe there is a bug in the script?

vaikas: nah, you're right as usual. Been a long day. Sorry for the noise :)

PR #748, opened 2021-08-23
Add g3n-github-robot account to the knative and knative-sandbox org

chizhg:
We have an internal tool at Google called G3N to enable sending notifications to Googlers when issues and pull requests need  their attention, and it's based on GitHub webhooks. 

One feature it supports is also sending notifications to people in GitHub teams, but since the GitHub teams for Knative are only visible to Knative members, we need to create a bot account, add it to the Knative org, and use its token to query the GitHub APIs.

`g3n-github-robot` is the bot account we created, and we will only use its token to query the team members with the GitHub APIs.

### DIFF
--- a/peribolos/knative-sandbox.yaml
+++ b/peribolos/knative-sandbox.yaml
@@ -94,6 +94,7 @@ orgs:
     - fj
     - flavorjones
     - frnksgr
+    - g3n-github-robot
     - garron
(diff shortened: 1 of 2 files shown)


evankanderson:
I didn't realize team memberships aren't public!

/lgtm
/approve

PR #758, opened 2021-08-27
Implement 731

### LINKED ISSUE #731: PROCESS CHANGE: Introduce end-user seats in steering committee

vaikas:
Fixes #731 

/hold

Just holding until the discussion in the #731 gets resolved, this is mirroring the plan there so it gets preserved.

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/STEERING-COMMITTEE.md
+++ b/STEERING-COMMITTEE.md
@@ -142,17 +142,14 @@ There are two ways to raise issues to the steering committee for decision:
 
 ## Composition
 
-The steering committee has 5 seats. These five seats are Contributing seats,
-open to any project contributor. In the future, the steering committe would like
-to add two End-User seats, open only to active end-user community members, once
-the selection criteria have been established. See
-[candidate eligibility](#candidate-eligibility) for definitions of these
-categories.
+The steering committee has 5 seats. Of these five seats, four (4) are
+Contributing seats, open to any project contributor, and one (1) seat is held
+by a KSC appointed End-User seat. End-User seat, is open to active end-user
+community members whose company is listed in the [ADOPTERS](./ADOPTERS.md) file.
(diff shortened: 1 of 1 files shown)


bsnchan:
Nice! Thanks for writing this up

thisisnotapril (approved):

PR #768, opened 2021-09-09
fix issue 92

### LINKED ISSUE #92: Reconcile knative-milestone-maintainers with project roles
Currently the Knative Milestone Maintainers group is used by Prow to gate access to `/milestone` commands.  I'd like guidance on how this aligns with our project roles, but it is also plausible that we should simply be more open and allow org members use this until it becomes a problem.

vaikas:
Remove the knative-milestone-maintainers role. Fixes #92 

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


vaikas:
/hold
Making sure that this won't cause grief with prow.

vaikas:
This seems troubling:
https://github.com/knative/test-infra/blob/main/guides/manual_release.md#option-1-via-prow-ui

But, I don't see anything in the test-infra repo that would indicate where this is configured. Would one of @chizhg @n3wscott mind taking a look?

dprotaso:
It's here: https://github.com/GoogleCloudPlatform/oss-test-infra/pull/937

vaikas:
I always forget about that repo 🤯 Thank you so much @dprotaso !!!

Given that TOC + Release leads are there, I reckon that we should drop the milestone maintainers from that list that you point to, then merge this.

vaikas:
/unhold
@dprotaso PTAL :)

PR #769, opened 2021-09-10
Removes bsnchan from steering

bsnchan:
Going on leave :'( - It's been a blast!

I suspect SC will leave the seat empty until the next steering elections in early October so the seat will remain empty for the timebeing.


vaikas:
I'll merge this and then I'll add @bsnchan to the Emeritus in a follow-up PR just in case she's no longer looking at PR comments :) 
/lgtm
/approve

PR #784, opened 2021-09-28
Update versions tables again, it turns out we had 3

evankanderson:
Removed the two copies in the long principles doc so we don't need to change our principles all the time. Centralized to one (smaller) table, plus a historical appendix.


julz (commented):
/lgtm
/assign @dprotaso

PR #785, opened 2021-09-28
Rotate in new release leads

evankanderson:
Per https://github.com/knative/release, @julz and @bmo will be the release leads for the 2021-11-02 release, tentatively named 0.27.

/assign @julz @bmo
/assign @pmorie

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -69,7 +69,7 @@ aliases:
   - pierDipi
   - vaikas
   knative-admin:
-  - ZhiminXiang
+  - bmo
   - bsnchan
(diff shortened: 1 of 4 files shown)


OWNERS_ALIASES
```diff
@@ -69,7 +69,7 @@ aliases:
   - pierDipi
   - vaikas
   knative-admin:
-  - ZhiminXiang
+  - bmo
```
julz, line 72: looks like this should be @benmoss

julz (changes requested):
wrong GitHub handle for @benmoss 

/hold

PR #826, opened 2021-11-04
add naina and omer as approvers

omerbensaadon:
adding omer and naina as func approvers so we can edit labels


dprotaso:
> failed to configure knative-sandbox members: all team members/maintainers must also be org members: nainaz

Can you add nainz to the member list?

omerbensaadon:
yes

lance (commented):
/lgtm

PR #836, opened 2021-11-08
Change the number of elected seats in this election properly to two.

jberkus:
Add notes about the seperate process for the End User seat.

As discussed, while three seats will change in this election, only the two Contributing Seats will be elected via Elekto.  The remaining seat will be selected by the newly seated SC.  This clarifies both the README and the description.

/assign @geekygirldawn, @vaikas


vaikas:
Thanks for clarifying the language, much better!
/lgtm
/approve

Issue #827, opened 2021-11-04
Clarify K8s min version principle given the introduction of the maintenance mode


dprotaso:
K8s upstream has extended their support policy to 14 months. The last 2 months of this support window is 'maintenance mode' and [is described as](https://kubernetes.io/releases/patch-releases/#support-period):

> During the two-month maintenance mode period, Release Managers may cut additional maintenance releases to resolve:
>
> - CVEs (under the advisement of the Security Response Committee)
> - dependency issues (including base image updates)
> - critical core component issues
>
> At the end of the two-month maintenance mode period, the patch release series will be considered EOL (end of life) and cherry picks to the associated branch are to be closed soon afterwards.
>
> Note that the 28th of the month was chosen for maintenance mode and EOL target dates for simplicity (every month has it).

Our min K8s policy states we support whatever is supported upstream at the time of a Knative release. We should update the language to exclude the 'maintenance mode'

dprotaso:
We need to update the k8s versioning principles now that K8s has a maintenance mode window.

rhuss:
> Our min K8s policy states we support whatever is supported upstream at the time of a Knative release. We should update the language to exclude the 'maintenance mode'

Would it make sense to also include such a notion of a 'maintenance mode'? I mean, when a k8s enters maintenance mode it will be still  2 months around, which is for one future Knative version, too. I don't mind being more restrictive than k8s and that is probably the right thing to do. 

So +1 from my side but would love to hear some more opinions before merging it.

nainaz:
I would like to see us defining such maintenance mode too. 
But, I am probably not seeing the implications for such action. 

For example: Knative enters maintenance mode with Kubernetes

PR #838, opened 2021-11-09
Update our min K8s version principle to exclude releases in maintenance mode

dprotaso:
Fixes: https://github.com/knative/community/issues/827

### DIFF
--- a/mechanics/RELEASE-VERSIONING-PRINCIPLES.md
+++ b/mechanics/RELEASE-VERSIONING-PRINCIPLES.md
@@ -29,9 +29,10 @@ webhooks, bugs)._
 ### K8s minimum version principle:
 
 We (the community) support the range of upstream minor Kubernetes versions that
-are Kubernetes community supported at the time of the cut. Only the latest patch
-release of the minor version is tested for support. Support for versions prior
-to that are best effort, or by vendors.
+are Kubernetes community supported at the time of the cut. This excludes releases
+which have entered maintenance mode. Only the latest patch release of the minor 
+version is tested for support. Support for versions prior to that are best effort, 
+or by vendors.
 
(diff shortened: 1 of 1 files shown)


evankanderson:
Policy reference: https://kubernetes.io/releases/patch-releases/#support-period

The rationale is presumedly to reduce the size of our testing matrix, and we don't expect that these releases will change much?

Or is the rationale to get access to new Kubernetes features 2 months earlier?

dprotaso:
I view the maintenance window as an extra two month buffer to transition to a K8s version in the 'standard support period'. 

We should have an available Knative release that supports that 'newer' version.  We generally 'support' newer K8s version but we don't stress them the same way we do our target min version - ie. running prow scale tests etc.

dprotaso:
@rhuss posted https://github.com/knative/community/issues/827#issuecomment-970041524

> > [dave] Our min K8s policy states we support whatever is supported upstream at the time of a Knative release. We should update the language to exclude the 'maintenance mode'
> 
> Would it make sense to also include such a notion of a 'maintenance mode'? I mean, when a k8s enters maintenance mode it will be still 2 months around, which is for one future Knative version, too. I don't mind being more restrictive than k8s and that is probably the right thing to do.
> 
> So +1 from my side but would love to hear some more opinions before merging it.

evankanderson:
/approve

Discussed this in the TOC meeting at Nov 18; we might also want to say "we (the community) will _test_" (rather than "support").

dprotaso:
> we might also want to say "we (the community) will _test_" (rather than "support").

'support' is qualified later in the paragraph

>  Only the latest patch release of the minor version is tested for support. Support for versions prior to that are best effort, or by vendors.

markusthoemmes:
LGTM

PR #840, opened 2021-11-10
address an identity crisis

lance:
Everywhere except GitHub, my user ID is lball - but in GitHub it's lance. In the SC elections candidate .md file for myself, I put my ID as lball and the file is named candidate-lball.md. This is an embarrassing mistake - but hopefully, this change will fix the entry in the election software.


jberkus:
Unfortunately, as voting has already started, we can't change your candidate ID without the votes being lost.  The candidate ID doesn't actually have to match Github, it just has to match itself; we ask for GitHubID to ensure that candidate IDs are unique and work as URLs.

My suggestion would be an update to your description only linking your actual github profile.

PR #846, opened 2021-11-16
Add verify-codegen workflow

benmoss:
Do it manually because the only thing knobots supports is a denylist, and it's too annoying to bother excluding every action except this single one.

If https://github.com/knative-sandbox/knobots/pull/99 was complete/merged this would be a lot easier, but unfortunately that's a dramatic refactoring that is now out of sync.

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -1,0 +1,73 @@
+# Copyright 2020 The Knative Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file is **********NOT********** automagically synced here by knobots
+# because its too hard to opt-in to a single action. See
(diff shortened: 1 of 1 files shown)


.github/workflows/verify.yaml
```diff
@@ -0,0 +1,73 @@
...
+          sed ':begin;$!N;s/\n/%0A/;tbegin'
+        }
+        if [[ -z "$(git status --porcelain)" ]]; then
+            echo "${{ github.repository }} up to date."
+        else
+            # Print it both ways because sometimes we haven't modified the file (e.g. zz_generated),
+            # and sometimes we have (e.g. configmap checksum).
+            echo "Found diffs in: $(git diff-index --name-only HEAD --)"
+            for x in $(git diff-index --name-only HEAD --); do
+                echo "::error file=$x::Please run ./hack/update-codegen.sh.%0A$(git diff $x | urlencode)"
```
rhuss, line 69: Is the note "Please run ..." not a bit repetitive, if it is printed out in the end anyway ?

rhuss:
> Do it manually because the only thing knobots supports is a denylist, and it's too annoying to bother excluding every action except this single one.

+1 for not fiddling around with knobots only for this single action.

.github/workflows/verify.yaml
benmoss: This is 99% copypasta from https://github.com/knative/.github/blob/e7a2b3c0f5e263e994e6ff9c9b40fb998d333040/workflow-templates/knative-verify.yaml

I just changed the header and got rid of the conditional in the bash (since this repo definitely has a hack/update-codegen)

benmoss: I think this part also is what makes it print annotated errors on the lines of code that don't match in the diff view, whereas the `echo` at the end just outputs to the action stdout

PR #872, opened 2021-12-15
Update release schedule

nak3:
As per title, 1.1 was released so update the schedule.

/cc @dprotaso @matzew


rhuss (commented):
/approve
/lgtm

PR #883, opened 2022-01-07
Clean up language around TOC vacancies

jberkus:
Per discussion with SC.  Specifically:
- replace "most votes" with "most preferred" to reflect the actual election mechanism
- make it clear that any replacement candidate serves the replaced members' term, regardless of how they were selected

please review: @geekygirldawn @vaikas


vaikas:
/lgtm
/approve

Thanks!

PR #888, opened 2022-01-11
Create candidate-sebgoa.md

sebgoa:
submission of candidacy for EU seat in the SC


sebgoa:
@dprotaso here is a fresh one, thanks

PR #889, opened 2022-01-11
Update releease schedule

evankanderson:
I was reviewing a different schedule and figured since I was looking at calendars, I might as well extend this to end of year.


evankanderson:
CC @dprotaso 

I figured I'd send this cross-company just to be extra-kosher, not sure why I thought it mattered...

PR #895, opened 2022-01-14
Add murugappan nominaton for end user sc seat

itsmurugappan:
Adding my nomination for end-user SC seat


csantanapr:
We welcome your nomination

/approve
/lgtm

PR #910, opened 2022-01-27
Add cardil@ as productivity writer

chizhg:
Chris has contributed to Knative productivity since 2019, and he had made a lot of great achievements on improving Knative testing, specifically upgrade testing, conformance testing and the general testing infrastructure.

I believe he has met all the requirements from a long while ago, so nominating cardil@ as a productivity writer.

Also removing two Googlers that have left Knative.

/cc @vaikas @csantanapr 

FYI @cardil

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -149,13 +149,12 @@ aliases:
   - efiturri
   - evankanderson
   - gerardo-lc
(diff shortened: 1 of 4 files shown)


cardil:
Thanks a lot @chizhg for kind words!

PR #911, opened 2022-01-28
remove myself from rabbitmq

benmoss:
I'm not participating in eventing-rabbitmq anymore so just to avoid getting mentions from the team alias


rhuss (commented):
/approve
/lgtm

PR #914, opened 2022-01-28
Add TOC to approvers for OWNERS_ALIASES

### LINKED ISSUE #913: PROCESS CHANGE: TOC should be able to approve OWNERS_ALIASES
Once upon a time, we had individual `OWNERS` files scattered throughout our repos, with names in them and an occasional (unsynchronized) `OWNERS_ALIASES` file. Then we instituted Peribolos, which allowed us to standardize GitHub group management in a GitHub file. And we granted the TOC and Steering the ability to manage this file.

Later, we tried to make these groups work for approvals management, using GitHub's `CODEOWNERS` file. But it turns out that there are `OWNERS` features that we like, so I wrote a tool to generate `OWNERS_ALIASES` files and copy them to various repos.

However, the side effect of this was that we were generating `OWNERS` files for `knative/community`, where our governance documents live. At that time, the community was also having some trust issues, and so I didn't propose giving TOC the ability to update `knative/community`'s `OWNERS_ALIASES` file. The unfortunate side-effect was that TOC members (who had been able to manage group membership joins/departures) now needed to ping steering for each change that would touch `OWNERS_ALIASES`.

In steering discussion last week (20 Jan 2022), steering agreed that this wasn't really necessary, and that we should give the TOC ability to approve these `OWNERS_ALIASES` changes.

(I'll send a PR for this process change shortly)

evankanderson:
Fixes #913


evankanderson:
I'm wondering whether the OWNERS config will be okay/happy with both per-file and top-level OWNERS. If not, I'll add a per-file match on `.*`.

thisisnotapril:
lgtm!

PR #915, opened 2022-01-29
update release leads for 1.3

psschwei:
# Changes

Updates release leads for 1.3

/cc @carlisia @xtreme-sameer-vohra @dprotaso

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -66,6 +66,7 @@ aliases:
   - pierDipi
   - vaikas
   knative-admin:
+  - carlisia
   - csantanapr
   - dprotaso
   - duglin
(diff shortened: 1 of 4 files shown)


OWNERS_ALIASES
```diff
@@ -66,6 +66,7 @@ aliases:
   - pierDipi
   - vaikas
   knative-admin:
+  - carlisia
```
vaikas, line 69: Was this intentional? Didn't see it in the PR description.

vaikas:
/lgtm
/approve
/hold

Just added the hold if the addition to admin was not intentional. If it was, please remove the hold.

OWNERS_ALIASES
psschwei: That was done by the [update-codegen](https://github.com/knative/community/blob/main/hack/update-codegen.sh#L27-L28) script; I assume it's needed to do things like cut branches for the repos.

dprotaso: looks like this is done since the release team is a sub team of the admin group

PR #920, opened 2022-02-04
Adding mgencur to productivity reviewers

cardil:
# Changes

- Adding [mgencur](https://github.com/mgencur) to productivity reviewers.
- Adding missing Red Hat QE team members to Knative orgs

Martin has a long history of Knative contributions. He should be in productivity reviewers a long time ago.

/kind enhancement

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -151,7 +151,9 @@ aliases:
   - efiturri
   - evankanderson
   - gerardo-lc
+  - jrangelramos
   - kvmware
+  - mgencur
   - shinigambit
   productivity-wg-leads:
   - chizhg
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -151,7 +151,6 @@ aliases:
   - efiturri
   - evankanderson
   - gerardo-lc
-  - jrangelramos
   - kvmware
   - mgencur
   - shinigambit
--- a/peribolos/knative-sandbox.yaml
+++ b/peribolos/knative-sandbox.yaml
@@ -157,6 +157,7 @@ orgs:
     - liu-cong
     - lkingland
     - macruzbar
+    - maschmid
     - markfisher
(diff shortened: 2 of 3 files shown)


rhuss (commented):
/approve
/lgtm

cardil:
Looks like this doesn't work:

https://github.com/knative/community/blob/main/OWNERS#L10-L17

Otherwise, @rhuss should be able to approve.

rhuss:
cc @evankanderson looks like the wildcard expansion doesn't work. (Is it supposed to be a glob or a regexp ?)

rhuss (approved):

rhuss:
well, is it a regexp according to https://www.kubernetes.dev/docs/guide/owners/#filters

But could it be that you can't apply filters to top-level OWNERS and OWNERS_ALIASES files ?

rhuss:
ah, from the link above:

> WARNING: The approve plugin [does not currently respect filters](https://github.com/kubernetes/test-infra/issues/7690). Until that is fixed, filters should only be used for the labels key (as shown in the above example).

dprotaso:
yeah the owner file is misconfigured

OWNERS_ALIASES
```diff
@@ -152,6 +152,7 @@ aliases:
   - evankanderson
   - gerardo-lc
   - kvmware
+  - mgencur
```
chizhg, line 155: Just in case you missed some contexts - currently adding a person to the reviewers group does not have an effect, due to the issue described in https://github.com/knative/test-infra/issues/2942

rhuss:
@knative/steering-committee could you please approve this PR (as the TOC still can't do it) ?

@chizhg I see, that it technically does not really make an effect whether you are mentioned as reviewer or not, but let's at least try to track the reviewers in the file, so that in case we want to go back to an automatically assigned reviewer, we already have the configuration in place.

PR #922, opened 2022-02-04
update meeting notes link to point to the new document

dprotaso:


working-groups/WORKING-GROUPS.md
```diff
@@ -66,7 +66,7 @@ The current working groups are:
 - [Security](#security)
   <!-- TODO add charters for each group -->
 
-## API Core
+## Serving API
```
psschwei, line 69: This breaks the section link above (L56), may want to update up there too

psschwei (changes requested):

working-groups/WORKING-GROUPS.md
psschwei: I'd also move this down so that the list stays in alphabetical order

dprotaso: oh nice catch

psschwei (approved):
/lgtm

Issue #917, opened 2022-02-02
Clarifying end-user seat selection for the SC for future elections


rhuss:
First of all, great to have an end-user seat in the SC which brings in a new vendor-independent perspective ! And welcome @itsmurugappan, great to have you on board !

However, I think there are some flaws in the process of end-user seat selection/election, which slipped through and I think should be fixed for the future (we did this the first time, so errors are kind of natural):

* In https://github.com/knative/community/blob/main/mechanics/SC.md we still don't specify the criteria who can apply for an end-user seat and the wording is still forward looking:

> End-user seats are planned to enable employees of organizations that use but do not sell Knative-derived products. The mechanics of eligibility for end-user seats are still in progress; it is expected that end-user seats will add to the current Steering options to specifically include end-user perspectives in the decision-making process.

* In https://github.com/knative/community/blob/main/elections/2021-SC-EU/README.md#eligibility we say that every user from a company mentioned in https://github.com/knative/community/blob/main/ADOPTERS.MD is eligible for an end-user seat. But this file contains also vendors like Red Hat, VMware, Google Cloud, IBM, SAP, Triggermesh, ... which I consider to be _vendors_, not _end users_.

* I can't really find the place where it says, that the end-user seat is elected by the other 5 seats in the steering committee (but that could be also my inability to read docs properly ;-). Can please someone point me to the place where this is written? Maybe we can include this more prominently in the SC.md document.

cc @knative/steering-committee

csantanapr:
Thanks for opening this @rhuss I added to the SC agenda and project backlog.

I agree we need to update SC.md to make it clear.

About selecting the end user seat we have it documented in the election [doc](https://github.com/knative/community/blob/main/elections/2021-SC-EU/README.md#background)
In 2021 we introduced one end user seat to bring more diversity into the SC representation **which is to be elected by the SC once the SC election was completed.**

We should move this wording to the SC.md like you said to make it clear.

Good catch on vendors vs end-user companies.

One way to resolve this is to update the take in ADOPTERS.MD to add a column "Vendor or EndUser"  and explanation on what is considered a vendor vs an organization end user.

devguyio:
[...]

PR #926, opened 2022-02-14
Update EU seat

csantanapr:
Related #917 

This update covers 2 points from the issue
- Reflects that now we have End User representation by indicating which member holds the seat
- How are elections for End User seat


itsmurugappan (approved):
I assume the end-user seat qualification will be updated in a separate PR.

/lgtm

csantanapr:
>I assume the end-user seat qualification will be updated in a separate PR.
Yes, @pmorie is going to work on ADOPTERS.md and end-user qualification

PR #934, opened 2022-02-28
Add google-ospo-team as the owner of Knative orgs

chizhg:
We can remove all the Google-owned bot accounts after Knative becomes part of CNCF.

### DIFF
--- a/peribolos/knative-sandbox.yaml
+++ b/peribolos/knative-sandbox.yaml
@@ -15,6 +15,7 @@ orgs:
     - dprotaso
     - evankanderson
     - google-admin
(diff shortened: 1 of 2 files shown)


evankanderson:
I'm missing the context on why we're adding another owner to the org -- the inner comment suggests removing a bunch of bots, but this is adding a bot in the meantime.

chizhg:
Sorry for the delayed response. 

Missing this bot account triggered some internal alerts, but it's not something serious. I think we can live with it for now, and avoid adding more bot accounts. I'll close this pull request.

PR #942, opened 2022-03-03
Copy logo PDF guidelines, including font and color guides, into the community branding guidelines.

evankanderson:
Scott Nichols noticed at the TOC meeting today that the logo colors on the website were off. We should include the details of the logo in the branding guidelines.

PR #972, opened 2022-03-19
Fix logo PDF guidelines file

sisp:
# Changes

The logo PDF guidelines file was invalid, probably an accidentally wrong download of https://github.com/knative/docs/blob/main/docs/images/logo/style-guide.pdf in #942. I've downloaded the PDF again and replaced the broken file with the right one.

/kind cleanup

### DIFF
--- a/logo-guidelines.pdf
+++ b/logo-guidelines.pdf
@@ -1,1528 +1,2642 @@
-
-
-
-
-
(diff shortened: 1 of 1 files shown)


sisp:
Just noticed this is a duplicate of #967.

PR #977, opened 2022-03-24
Ahmed stepping down from eventing and eventing-kafka wg leads and approvers

devguyio:
As already announced [here](https://groups.google.com/g/knative-dev/c/IqlSNL7jvdo) , it's time to for me to return back to the observer seat. 

/kind cleanup
/kind removal

I also suggest that we shutdown the Eventing Kafka WG. Right now it has no leads and I'm not sure if @pierDipi or @matzew would be willing to keep it up. But I feel that's a followup after this PR.

/assign @evankanderson @dprotaso @rhuss @n3wscott

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -53,13 +53,11 @@ aliases:
   - antoineco
   - lberk
   eventing-wg-leads:
-  - devguyio
   - lionelvillard
   eventing-writers:
   - akashrv
   - aliok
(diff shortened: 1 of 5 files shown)


evankanderson (commented):
/lgtm
/approve

/assign @itsmurugappan 

The end of an era. Sad to see you go, Ahmed, have fun at your next gig!

PR #981, opened 2022-03-28
Add gab-satchi to eventing-rabbitmq approvers

gab-satchi:
# Changes
Add gab-satchi to eventing-rabbitmq approvers


gabo1208:
/retest
cc @dprotaso @kvmware

evankanderson:
Assuming that @gabo1208 's comment is an approval.

/lgtm
/approve

evankanderson:
(It looks like the presubmit is not getting the access token it needs to talk to the GitHub API, so it may take a little while for this to propagate.)

We probably also need a Steering approver, because per-file approvals don't actually work.

PR #983, opened 2022-03-29
Add caniszczyk and thelinuxfoundation as org OWNERS

### LINKED ISSUE #954: GitHub: ensure 'thelinuxfoundation' and 'caniszczyk' are added as org owners
GitHub: ensure 'thelinuxfoundation' and 'caniszczyk' are added as initial org owners, this helps us make sure we have continuity of GH ownership

evankanderson:
Fixes #954

I'm also removing `julz` from the admin memberships, and switching him to a normal member.

Note that I'm _not_ adding them to any teams right now (including the "admin" teams), because I think they may need to accept the org invite before being eligible to be added to teams, and Peribolos will complain if I try to do both steps at once.

I'm also not removing the `google` bots yet, but I expect that will probably happen one of these days.

### DIFF
--- a/peribolos/knative-sandbox.yaml
+++ b/peribolos/knative-sandbox.yaml
@@ -140,6 +140,7 @@ orgs:
     - jrangelramos
     - JRBANCEL
     - jskeet
+    - julz
     - jzelinskie
     - k4leung4
     - knative-automation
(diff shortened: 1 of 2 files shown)


csantanapr:
Thank you @evankanderson 
/lgtm

Issue #970, opened 2022-03-17
PROCESS CHANGE: the LICENSE file should include an SPDX license identifier and be an approved CNCF license


lance:
The `LICENSE` file for this repository does not include an SPDX identifier. See: https://api.github.com/repos/knative/community

```json
 "license": {
    "key": "other",
    "name": "Other",
    "spdx_id": "NOASSERTION",
    "url": null,
    "node_id": "MDc6TGljZW5zZTA="
  },
```

The `LICENSE` file includes two licenses. For the documentation, there is the "Creative Commons Attribution 4.0 International" license. For samples and other code, the "Apache 2.0" license is included. Per the CNCF, the Creative Commons license is, unfortunately, not approved.

All of this is causing this repository to score fairly poorly on the CNCF CLO Monitor for licenses https://clomonitor.io/projects/knative/knative#community_license. (There are problems in other areas as well, but this issue is just meant to deal with the license problems). The checks that the CLO Monitor makes for licenses are documented here: https://github.com/cncf/clomonitor/blob/main/docs/checks.md#license

We need to determine whether or not we can change the license for documentation from Creative Commons to an approved license, and to ideally have only one license covering all code and documentation.

/cc @knative/steering-committee

vaikas:
Previous discussion here:
https://github.com/knative/community/issues/517

jberkus:
Please file a ticket with the CNCF.  This is their technical issue to fix.

lance:
@vaikas regarding the SPDX ID, the CloudEvents org handles this by having the full LICENSE file in the root of the repo, and each file within the repo simply contains the SPDX ID.

dprotaso:
Sorta related can we now drop the license header in our code in favour of spdx identifiers?

lance:
> Sorta related can we now drop the license header in our code in favour of spdx identifiers?

@dprotaso I think it's fine - that's what the cloudevents repos do and they are also governed by cncf

csantanapr:
This is the feedback from cncf
>Jeffrey Sica
Monday 10:23 AM
In other repos we’ve seen a LICENSE file and a LICENSE-docs file which delineates what license applies to what. In addition, GitHub is working on being able to visualize when there’s more than one applicable license in the repo. 
There are some conversations about requiring all files to have SPDX headers so the licenses are explicitly stated rather than implicitly, but I don’t think we’re at the point yet where we’re going to ask projects to tackle that. 
Does that clarify things enough for KNative to move forward? 🙂 

[...]

PR #985, opened 2022-03-30
Update license file(s)

lance:
As part of the transition to CNCF, we now have a CLO Monitor tool which
tells us if our licenses are up to snuff. The license for this repository
is not.

https://clomonitor.io/projects/knative/knative#community_license)

This commit splits the license into two separate files, the LICENSE file
covers all code, and the LICENSE-docs file covers all docs, per CNCF
reccommendations. https://cncfservicedesk.atlassian.net/servicedesk/customer/portal/1/CNCFSD-1168

Fixes: https://github.com/knative/community/issues/970

/kind cncf-infra


lance:
> /hold @lance I'm concerned that this PR deletes the file `LICENSE` is intentional? My understanding was we were going to have 2 files `LICENSE` and `LICENSE-docs` Not having `LICENSE` affects github oss ranking

This PR does not delete the `LICENSE` file at all! :) It simply removes the Creative Commons section and puts it in a new `LICENSE-docs` file.

LICENSE-docs
```diff
@@ -0,0 +1,390 @@
+SPDX-License-Identifier: Apache-2.0
```
csantanapr, line 1: Is this a mistake?

csantanapr: Should this crative commons

lance: @csantanapr yes good catch. It should be in the main `LICENSE` file. Fixed.

PR #992, opened 2022-03-31
Remove skonto is not in org

kvmware:
unblocks https://github.com/knative-sandbox/knobots/pull/163

```
skonto
User is not a member of the org. User is not a collaborator. Satisfy at least one of these conditions to make the user trusted.
```


gabo1208:
You need to run ./hack/update-codegen.sh @kvmware :)

PR #993, opened 2022-03-31
Remove skonto, not in org

kvmware:
Unblocks https://github.com/knative-sandbox/knobots/pull/163

```
- skonto
-- User is not a member of the org. User is not a collaborator. Satisfy at least one of these conditions to make the user trusted.
```

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -7,7 +7,6 @@ aliases:
   autoscaling-reviewers:
   - nader-ziada
   - psschwei
-  - skonto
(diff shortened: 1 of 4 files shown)


dprotaso:
@skonto accepted the invite - we shouldn't need to remove him

skonto:
What needs to be done here? I have accepted the knative-sandbox invitation already.

dprotaso:
Running`/verify-owners` on blocked PRs works to remove the bad label

PR #997, opened 2022-04-01
Clean up steering GSuite memberships

evankanderson:
Dave was asking questions about these groups files, and they didn't really make sense to me, so I figured I'd do a quick cleanup pass (and I don't have approval for this directory).


vaikas:
/lgtm
/approve
Thanks @evankanderson

PR #998, opened 2022-04-01
Update productivity group membership

evankanderson:
Many of the `wg-productivity` members listed are not currently working on Knative. Sub in new TOC members to the list instead (though we might consider removing them until they need the access -- it's unclear what the emergency fix scenario might be without them having this access).

Add TOC as permitted to manage wg-productivity memberships.

### DIFF
--- a/groups/wg-productivity/OWNERS
+++ b/groups/wg-productivity/OWNERS
@@ -2,6 +2,7 @@
 
 approvers:
 - productivity-writers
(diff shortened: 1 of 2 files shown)


vaikas:
Thanks @evankanderson 
/lgtm
/approve

chaodaiG (commented):
/lgtm
/approve

Glad to know that I still have this power

PR #1004, opened 2022-04-07
Clarify end user appointment, vacancy and companies

### LINKED ISSUE #995: [Steering] Update documents to clarify SC End User vacancy process
The SC document has information on how to handle [vacancies](https://github.com/knative/community/blob/main/STEERING-COMMITTEE.md#vacancies) for elected members, but the end user is an appointed seat by the members of the steering committee.

>In the event of a resignation or other loss of an elected SC member, the candidate with the next most votes from the previous election will be offered the seat, provided that person otherwise qualifies to join the SC. This process will continue until the seat is filled.
In case this fails to fill the seat, a special election for that position will be held as soon as possible, unless the regular SC election is less than 7 weeks away. Eligible voters from the most recent election will vote in the special election. Eligibility will not be redetermined at the time of the special election. Any replacement SC member will serve out the remainder of the term for the person they are replacing, regardless of the length of that remainder.

We should update the document to clarify how the seat replacement for end user seat would work since "special election" is not applicable

Also to update places where we use the term "selected" or "elected" when referring to end user seat, we should use "appointed" 
like in the end user candidate [document](https://github.com/knative/community/blob/main/elections/2021-SC-EU/README.md)

cc @geekygirldawn @knative/steering-committee

csantanapr:
Fixes #995 #917

### DIFF
--- a/ADOPTERS.MD
+++ b/ADOPTERS.MD
@@ -8,26 +8,28 @@ If you would like to be included in this table, please submit a comment to [this
 
 | Organization                      | Product/Project Name                                                   | Status                   | More Information                                                     |
 | --------------------------------- | ---------------------------------------------------------------------- | ------------------------ | -------------------------------------------------------------------- |
-| Alibaba Cloud                     | Knative on Alibaba Cloud                                               | Production               | [link](https://cs.console.aliyun.com/)                               |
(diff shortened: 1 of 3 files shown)


vaikas (commented):
Just some little nits.

csantanapr:
@vaikas PTAL when you have a chance, thanks for the review

vaikas:
/lgtm
/approve

Thanks for doing this!

PR #1008, opened 2022-04-07
OWNERS filtering doesn't work for approvals - use symlinks instead

dprotaso:
Related issue: https://github.com/kubernetes/test-infra/issues/7690


dprotaso:
The knobots job will need updates

https://github.com/knative-sandbox/knobots/blob/main/actions/update-community/entrypoint.sh

dprotaso:
cc @evankanderson

hack/update-codegen.sh
```diff
@@ -24,8 +24,8 @@ function make_aliases() {
   # self-contained and we don't release them.
   ( cd mechanics/tools/gen-aliases; go build . )
   # Auto-generate OWNERS_ALIASES from peribolos configs
-  ./mechanics/tools/gen-aliases/gen-aliases knative peribolos/knative.yaml OWNERS_ALIASES
-  ./mechanics/tools/gen-aliases/gen-aliases knative-sandbox peribolos/knative-sandbox.yaml knative-sandbox-OWNERS_ALIASES
+  ./mechanics/tools/gen-aliases/gen-aliases knative peribolos/knative.yaml peribolos/knative-OWNERS_ALIASES
+  ./mechanics/tools/gen-aliases/gen-aliases knative-sandbox peribolos/knative-sandbox.yaml peribolos/knative-sandbox-OWNERS_ALIASES
```
vaikas, line 28: do we need an entry somewhere for:
peribolos/knative-sandbox-OWNERS_ALIASES

vaikas: Wouldn't let me comment on the entry in OWNERS_ALIASES file so just added it here.

evankanderson: I think this has entries for generating both files.

evankanderson:
> I'm guessing we're going to copy the OWNERS_ALIASES to the root in the same way we do for all the other repos?

Oh, I missed the first symlink. I'm assuming prow follows symlinks?

dprotaso:
> Oh, I missed the first symlink. I'm assuming prow follows symlinks?

Yes - that's the assumption

dprotaso:
I asked in k8s #prow : https://kubernetes.slack.com/archives/CDECRSC5U/p1649351219095019?thread_ts=1649349770.252429&cid=CDECRSC5U

hack/update-codegen.sh
dprotaso: I didn't include it because I was thinking of just updating the sync actions here:

https://github.com/knative-sandbox/knobots/blob/054b91b98bca48e7b42227e31f0590269b3a794d/actions/update-community/entrypoint.sh#L31-L35

dprotaso:
follow up: https://github.com/knative-sandbox/knobots/pull/183

PR #1010, opened 2022-04-07
drop filter since it applies only to labelling

dprotaso:
Forgot to remove the filter in https://github.com/knative/community/pull/1008


dprotaso:
bump @vaikas this is good now

PR #1012, opened 2022-04-11
Remove aryann from the knative peribolos config

chizhg:
The GitHub user updated their username and caused peribolos to fail, see the context in https://github.com/knative/test-infra/issues/3247.

For now removing them to unblock peribolos.

/cc @dprotaso @evankanderson


evankanderson (commented):
/lgtm
/approve

Issue #968, opened 2022-03-15
PROCESS CHANGE: CoC violations reporting and enforcement improvements/clarifications


lance:
Users currently have the ability to report Code of Conduct violations and concerns to code-of-conduct@knative.team. This is documented in https://github.com/knative/community/blob/main/CODE-OF-CONDUCT.md#enforcement. It should be stated clearly who is behind that email address (it is the @knative/steering-committee). In addition, there should be a mechanism to report Code of Conduct violations by Steering Committee members through another CNCF enforcement channel.

* The community benefits by having an anonymous means of reporting CoC violations by any member of the community, including those charged with enforcing it.

If the police police police the police, who polices the police police?! 😆

PR #1016, opened 2022-04-19
docs: remove old CoC info from README.md

lance:
See also https://github.com/knative/community/issues/968#issuecomment-1102855282

Do we continue to monitor the knative email for CoC reporting? Or is this all in CNCF hands now?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,3 @@
-_Important_. Before proceeding, please review the Knative community
-[Code of Conduct](./CODE-OF-CONDUCT.md).
-
-If you any have questions or concerns, please contact the authors at
-knative-code-of-conduct@googlegroups.com.
-
 ## Welcome!
 
 Welcome to the Knative community!


vaikas:
I see the link to COC has been updated, so this LGTM. Just adding a hold so others can chime in just in case they have concerns (remove at will if there are none).
/lgtm
/approve
/hold

lance:
@vaikas what I'm really curious about, though, is whether we will continue to use the code-of-conduct@knative.team email address to handle reporting, as it used to be documented here: https://github.com/knative/community/blob/cfd6cdc76616e917ece51d7aa69fea64a3419d9e/CODE-OF-CONDUCT.md?plain=1#L62-L73

#968 is obsolete/irrelevant if _all_ CoC violations go through the CNCF reporting structure. But I think it's perhaps better to restore some of the old enforcement text in our CoC, so that the first line of defense is Knative SC, but if that's insufficient or sensitive, issues can be elevated to CNCF enforcement. WDYT?

vaikas:
Yeah, understood :) I went to look for what it says earlier and noticed it's a redirect, and indeed it's a pointer to the CNCF documentation. I don't see a point in having a separate email for reporting to Knative ones if we point to CNCF that has it's own set of reporting instructions?

```
For other projects, please contact the CNCF staff via [conduct@cncf.io](mailto:conduct@cncf.io). You can expect a response within three business days.
```

It explicitly lists kubernetes as it's own project, but I don't see any others. Are there other projects that handle their own COC despite being in the CNCF?

I think since we're adhering to CNCF COC, then having them (assuming they are willing to do it) sounds like it would be probably holistically a more consistent way to handle this across the whole CNCF eco system?

lance:
I took a look at buildpacks/pack to see what they are doing and they are doing their own thing entirely (which I find odd). https://github.com/buildpacks/.github/blob/main/CODE_OF_CONDUCT.md

vaikas:
Unless there's a really good reason that I don't understand which is entirely possible, I'd rather have us adhere _and_ use the existing processes for dealing with violations. Seems simpler, more consistent, and don't have to worry about things diverging.

lance:
> Unless there's a really good reason that I don't understand which is entirely possible, I'd rather have us adhere _and_ use the existing processes for dealing with violations. Seems simpler, more consistent, and don't have to worry about things diverging.

I'm OK with that. Maybe we can formalize that decision at the SC call tomorrow and then I will close [this](https://github.com/knative/community/issues/968).

lance:
@vaikas updated per our latest understanding

vaikas:
Sounds good :) That was just like my opinion man :)

PR #1023, opened 2022-04-21
Add hh to Knative GitHub Org as a Member

hh:
See https://knative.slack.com/archives/CCSNR4FCH/p1650581831685649?thread_ts=1649809077.370369&cid=CCSNR4FCH

### DIFF
--- a/peribolos/knative.yaml
+++ b/peribolos/knative.yaml
@@ -211,6 +211,7 @@ orgs:
     - hatofmonkeys
     - Henrike42
     - hev
+    - hh
     - hilalguler-dev
     - hisarbalik
     - holgerkoser


csantanapr:
Hi @hh 
the test is failing because you need to regenerate owner alias file, please run this script https://github.com/knative/community/blob/main/hack/update-codegen.sh from the root directory

evankanderson:
The actual error is:

```
Configuration failed: failed to configure knative members: users in both roles: hh
```

Do you need to be a GitHub admin, or just an org member?

evankanderson:
You also need to join the org (and accept the invite) before we can add you to any of the Working Group GitHub groups -- a non-org member can't be a member of a team, and Peribolos backs off it's dry-run as soon as it sees a single error from GitHub.

PR #1024, opened 2022-04-21
Add hh to Knative-Sandbox GitHub Org as a member

hh:
See https://knative.slack.com/archives/CCSNR4FCH/p1650581849509699?thread_ts=1649809077.370369&cid=CCSNR4FCH


csantanapr:
Hi @hh 
the test is failing because you need to regenerate owner alias file, please run this script https://github.com/knative/community/blob/main/hack/update-codegen.sh from the root directory

PR #1028, opened 2022-04-26
Add cncf-ci account as admin

BobyMCbobs:
Slack conversation:
https://knative.slack.com/archives/CCSNR4FCH/p1651014165202289?thread_ts=1651010158.667779&cid=CCSNR4FCH


BobyMCbobs:
Thank you for approving @dprotaso.

Are these errors 
https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_community/1028/pull-knative-peribolos/1519121208333832192
meaning that a username can only exist in either member or admin?

dprotaso:
sorry - read the message - yeah you'll need to be a member _and_ admin

BobyMCbobs:
> sorry - read the message - yeah you'll need to be a member _and_ admin. so add to both lists

I'm seeing the error
```
... failed to configure knative members: users in both roles: cncf-ci ...
```

Gonna try to remove from members and keep in admin, @dprotaso.

BobyMCbobs:
Ah, that seemed to do it :rocket: 

Does it look fine to you, @dprotaso/

dprotaso:
ok - looks like I was wrong and you're either in the `admin` list or `members` but not both

PR #1030, opened 2022-04-27
Add OWNERS file field to candidates template

jberkus:
... per discussion with Steering on Slack.

This elections configuration update does two things:

1. Configures "affiliations" and "owners" fields as the visible fields on candidate profiles
2. Adds an "owners" field to the candidate profile

Reason: being named as an Approver in one or more Owners files is a requirement to run for TOC.  Since there is no master index of Owners files for Knative, EOs and voters need this info to validate candidate eligibility.


jberkus:
tag @geekygirldawn

PR #1034, opened 2022-05-04
Merge Scaling and Serving API Working Groups

dprotaso:
Mentioned the merger during the ToC Update April 28th.

This drops the `autoscaling-*` GitHub teams. All the `autoscaling-reviewers` are already present in a `serving-reviewers`. Likewise for the `*-writers` groups.

I created a `serving-wg-leads` which is the rename for `api-core-wg-leads`. Rename will have to happen once this rolls out and I update the appropriate `OWNER` files.

/hold
/assign @knative/technical-oversight-committee


working-groups/WORKING-GROUPS.md
```diff
@@ -297,33 +300,6 @@ performance/scale/load testing infrastructure
...
-| -------------------------------------------------------------- | --------------- | ------- | --------------------------------------------------- |
-| <img width="30px" src="https://github.com/julz.png">           | Julian Friedman | IBM     | [julz](https://github.com/julz)                     |
-
-| &nbsp;                                                         | Emeritus Leads  | Profile                                             | Duration  |
-| -------------------------------------------------------------- | --------------- | --------------------------------------------------- | --------- |
-| <img width="30px" src="https://github.com/markusthoemmes.png"> | Markus Thömmes  | [markusthoemmes](https://github.com/markusthoemmes) | 2019-2021 |
-| <img width="30px" src="https://github.com/vagababov.png">      | Victor Agababov | [vagababov](https://github.com/vagababov)           | 2019-2021 |
-| <img width="30px" src="https://github.com/josephburnett.png">  | Joseph Burnett  | [josephburnett](https://github.com/josephburnett)   | 2018-2019 |
-
```
evankanderson, line 326: Move this to "emeritus" or otherwise discontinued, rather than just deleting?

evankanderson (commented):
I'm not quite sure why peribolos is mad, but the change overall makes sense to me.

working-groups/WORKING-GROUPS.md
dprotaso: I moved it under 'Serving' with designated subgroups.

dprotaso:
We haven't added a new group in a while and peribolos now requires slugs for teams. So I'll have to retroactively add it

peribolos/knative-sandbox.yaml
```diff
@@ -803,14 +803,14 @@ orgs:
           reconciler-test: write
         teams:
           API Core WG Leads:
-            description: The Working Group leads for Serving API Core
+            description: The Working Group leads for Serving
             members:
             - dprotaso
             privacy: closed
```
evankanderson, line 809: I'm confused why we have both `api core wg leads` and `serving wg leads`

Is this an intermediate state until we get all the repos switched over to the new name?

dprotaso: Yup - exactly

psschwei (commented):
/lgtm

PR #1035, opened 2022-05-04
Adding Scott as 2022 TOC candidate

n3wscott:
Hello! I am Scott.

I have been contributing to Knative since initial public launch. In that time I
held the role of Sources Working Group Lead, as well leading the Productivity
Working Group during a time of uncertainty. I was a core contributor to Eventing
and was also a key member reconciler framework and generator code. I am a
maintainer on the CloudEvents Go SDK, a foundational component of Eventing. I
have most recently been serving on the Knative TOC, and helped in the planning
committee with our very first KnCon Day Zero event (!!!HYPE!!!).

I care deeply about the direction and future of Knative as a project. It is
vitally important to continue the work we are doing in Knative to level the
playing field for all of the Cloud Native landscape to make Cloud Computing more
accessible to those who want to provide business value on-top of an already
complex landscape. I want to provide a voice for Knative from the prospective of
a small business whose mission is not directly related to Knative, but Knative is
that enabling technology to let us move fast in the real world.

### DIFF
--- a/elections/2022-TOC/canidate-n3wscott.md
+++ b/elections/2022-TOC/canidate-n3wscott.md
@@ -1,0 +1,24 @@
+-------------------------------------------------------------
+name: Scott Nichols
+ID: n3wscott
+
+info:
+  - affiliation: Chainguard, Inc.
+  - owners: Current TOC
+-------------------------------------------------------------
+
+I have been contributing to Knative since initial public launch. In that time I
+held the role of Sources Working Group Lead, as well leading the Productivity
+Working Group during a time of uncertainty. I was a core contributor to Eventing
+and was also a key member reconciler framework and generator code. I am a
+maintainer on the CloudEvents Go SDK, a foundational component of Eventing. I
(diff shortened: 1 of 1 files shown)


geekygirldawn:
/hold 
Putting a hold to make sure this doesn't get accidentally merged until after the nomination period closes so that people can continue to comment / endorse on this PR.

I know @n3wscott *should* be in the TOC OWNERS file, but I think maybe you are missing? It's also possible I'm looking in the wrong place :) 
https://github.com/knative/community/blob/main/peribolos/knative.yaml
https://github.com/knative/community/blob/main/peribolos/knative-OWNERS_ALIASES

Regardless, based on the fact that you're already on the TOC means you meet the criteria, but can we also clean up the OWNERS or let me know I'm looking in the wrong place.

n3wscott:
I was added off cycle, perhaps that PR never got made to add me to the peribolos files, which explains some of the issues I have been having with approvals 😅

PR #1051, opened 2022-05-06
chore: add original boson contributors to knative

lance:

Adds @lkingland @matejvasek and @zroubalik to knative org


evankanderson (commented):
It looks like these three are already in `knative-sandbox`.

/lgtm

PR #1052, opened 2022-05-06
[wip] use slug-format for team names

dprotaso:
It seems like updates to the peribolos tooling don't like team names that aren't in slug format - ie. `API WG Leads` should be `api-wg-leads`

See discussion: https://kubernetes.slack.com/archives/CDECRSC5U/p1651764427799469


dprotaso:
Talking it over - this isn't what we want. Peribolos should work with teams in the prior format

PR #1060, opened 2022-05-11
add myself as async-component approver

AngeloDanducci:
Now that the async-component is in Beta and may get some more helping hands with the upcoming KnativeCon talk I think adding myself as approver for the async-component makes sense 🎉


maximilien (approved):
Thanks for all the great contributions @AngeloDanducci 

/approve
/lgtm

PR #1069, opened 2022-05-19
Proposing that SC, TOC, and Trademark Committee members should be eligible to vote.

geekygirldawn:
This PR adding that SC, TOC, and Trademark Committee members are eligible to vote regardless of the number of contributions for TOC and SC elections. Every year, we have several committee members who make significant contributions, but who don't hit the 50 devstats contributions. Rather than making them file exception requests for every election, I think we should just make them eligible by default.

This change requires SC approval. 

/hold

### DIFF
--- a/mechanics/SC.md
+++ b/mechanics/SC.md
@@ -59,6 +59,9 @@ year but is not captured in the stats.knative.dev dashboard, they will be able
 to submit an exception form to the steering committee who will then review and
 determine whether this member should be marked as an exception.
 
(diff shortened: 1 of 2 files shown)


jberkus:
+1 from the other EC

lance:
@knative/steering-committee we have neglected to discuss this in our last two meetings. I think it makes a lot of sense. Keeping the `/hold` until others chime in, but this lgtm.

/lgtm

PR #1071, opened 2022-05-23
Add GH Labels

upodroid:
# Changes

I need some labels to prioritise some work and few area specific labels for test-infra

/cc @dprotaso @kvmware

### DIFF
--- a/label_sync/labels.yaml
+++ b/label_sync/labels.yaml
@@ -265,6 +265,41 @@ default:
       prowPlugin: size
       addedBy: prow
 
+    #############################################################################
+    # Prioritisation labels borrowed from Kubernetes
+    #############################################################################
+
+    - color: fef2c0
+      description: Lowest priority. Possibly useful, but not yet enough support to actually get it done. # These are mostly place-holders for potentially good ideas, so that they don't get completely forgotten, and can be referenced /deduped every time they come up.
+      name: priority/awaiting-more-evidence
+      target: both
+      prowPlugin: label
+      addedBy: anyone
+    - color: fbca04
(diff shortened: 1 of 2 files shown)


label_sync/labels.yaml
```diff
@@ -265,6 +265,41 @@ default:
...
+      name: priority/important-longterm
+      target: both
+      prowPlugin: label
+      addedBy: anyone
+    - color: eb6420
+      description: Must be staffed and worked on either currently, or very soon, ideally in time for the next release.
+      name: priority/important-soon
+      target: both
+      prowPlugin: label
+      addedBy: anyone
```
kvmware, line 301: These seem to be what one would use the project board for which sorta captures these states already.

label_sync/labels.yaml
```diff
@@ -276,6 +311,31 @@ repos:
...
+    - color: 0052cc
+      name: area/tools
+      description: Issues or PRs related to internal tools found at /tools
+      target: issues
+      addedBy: anyone
+    - color: 0052cc
+      name: area/images
+      description: Issues or PRs related to testing docker images found at /images
+      target: issues
+      addedBy: anyone
```
kvmware, line 338: :+1:

label_sync/labels.yaml
upodroid: It doesn't today.

https://github.com/orgs/knative/projects/40/views/1

For example, we have 10 items ready to work but they don't have the priority labels so everyone doesn't know which one is important, work that we have deffered

kvmware: image
Notice you can even title the priority field as you like

upodroid: I didn't see that column. Cool

dprotaso:
@upodroid do you need this PR anymore (since you can add custom columns to GitHub Projects )

upodroid:
I still think we need it. The columns aren't as good as labels as the fields are write-in, the column doesn't appear in the issues list or in the specific issue's page.

PR #1098, opened 2022-06-09
shift WG updates by a week by request from DUX WG

csantanapr:
DUX WG requesting an extension to present to TOC


dprotaso:
Can you also update the reminds in #wg-leads channel?

PR #1104, opened 2022-06-30
chizhg steps down from Productivity WG lead and leaves Knative

chizhg:
/cc @evankanderson @kvmware


kvmware:
/lgtm
If we don't merge this will you stay with us @chizhg ? :P
Thanks again for all your work and help. I will miss you!

evankanderson (commented):
/approve
/lgtm

PR #1109, opened 2022-07-05
chore: remove func from knative org in peribolos

lance:
We have not found bandwidth to move the repo yet, and the fact that the
teams/groups are out of sync with reality needs to be addressed in the
meantime. This commit shifts ownership from the client to the functions
working group in knative-sandbox and introduces working group leads for
the two function repositories (kn-plugin-func and func-tastic). And incidentally
removes Omer from the committers list.

### DIFF
--- a/peribolos/knative-sandbox-OWNERS_ALIASES
+++ b/peribolos/knative-sandbox-OWNERS_ALIASES
@@ -114,9 +114,18 @@ aliases:
   - lkingland
   - matejvasek
   - nainaz
-  - omerbensaadon
   - salaboy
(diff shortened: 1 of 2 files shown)


kvmware:
/retest
https://github.com/knative/community/pull/1136 should ensure you adding the right options. So might want to wait for that to land first. Also, feel free to add the members to the team.

lance:
@kvmware whew! It's finally clean. Thanks for your assistance here.

zroubalik (approved):
/lgtm
/approve

PR #1117, opened 2022-07-12
Fixed / updated a few links

geekygirldawn:
Fixed a link to Elekto and changed links to previous elections to point to the elections folder to avoid having to change those links every year.

/assign @csantanapr @vaikas


vaikas:
Thank you!
/lgtm
/approve

PR #1124, opened 2022-07-14
process-change: loosen requirements for approvers

lance:
The approver role can be difficult to assign for repositories with lower
velocity or for those with a small number of (e.g. 2) primary
contributors, where the contributor seeking the approval role is making
most of the code contributions. This change attempts to address that
situation by removing a hard number requirement, and allowing for
individual contributions to be considered as well. Finally, instead of
simply suggesting that the contributor be in the OWNERS file, this
change strengthens that as a requirement.

### DIFF
--- a/ROLES.md
+++ b/ROLES.md
@@ -188,16 +188,9 @@ approver in an OWNERS file:
 
 - Reviewer of the codebase for at least 3 months.
 
-- Primary reviewer for at least 10 substantial PRs to the codebase.
+- Should have reviewed or contributed to substantial PRs to the codebase.
 
-  - One path for getting the necessary reviews is to add yourself to the
-    `reviewers` section of the OWNERS file. Note that this does not give you any
-    additional privileges. By having yourself listed in this section in OWNERS
-    file means that you will get PRs assigned to you for code review. Getting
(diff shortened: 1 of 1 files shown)


psschwei:
If we're doing this for approvers, should we also consider doing it for WG leads?

evankanderson (commented):
LGTM

(non-binding, not intended to trigger submit)

vaikas:
/lgtm
/approve
/hold
Adding a hold in case folks have objections, feel free to remove.

dprotaso:
With the changes - not sure if you want to remove this clause

```
Please note that repos in the knative-sandbox GitHub org are a reflection of the WG leads. These leads are free to use their judgement to set the bar for what is required to become an approver and/or owner for knative-sandbox repos they are responsible for.
```

lance:
> With the changes - not sure if you want to remove this clause
> 
> > Please note that repos in the knative-sandbox GitHub org are a reflection of the WG leads. These leads are free to use their judgement to set the bar for what is required to become an approver and/or owner for knative-sandbox repos they are responsible for.


@dprotaso I think that section is still valid with these changes. Seems so anyway.

zroubalik (approved):
/lgtm
/approve

PR #1136, opened 2022-07-27
Add peribolos schema validation workflow

kvmware:
This will catch errors that our current presubmit job doesn't without blowing up when adding a team. Additionally, this is a 10s run process vs the current 3min job.


kvmware:
@dprotaso

kvmware:
Fixes: https://github.com/knative/test-infra/issues/3384

dprotaso (approved):
/lgtm
/approve
/hold'ing for the nit

FYI: I'm a bit mixed on using python - it requires new contributors to learn python in addition to go. But this change seems simple enough.

dprotaso:
lol whoops

I guess you can't type `/hold'ing`

PR #1140, opened 2022-08-02
Fix Release Dates

dprotaso:
- Release dates now match what's in the GCal
- I adjusted the K8s min version for 1.7 to be 1.22 (from 1.23) - since maintenance mode for 1.22 is the week after 8/28 and 1.23 isn't in the GKE regular channel
- Changed the sort order (descending) of the historical table


evankanderson (commented):
/lgtm
/approve

Issue #1096, opened 2022-06-08
PROCESS CHANGE: Extend Knative Release Cycle Period


n3wscott:
Wanted to bring up the question of changing the Knative Release cycle period. At the moment we are releasing fairly trivial releases every 6 weeks, which points to a fact that we are releasing too often for the lifecycle of the project. 

There are other life-cycle stage features we might want to look at, like alpha/beta/GA releases of single release versions. 

Let's start the conversation and then follow up with an AI to have someone lead writing a document to work the details out for doing less releasing, making the releases a bit more meaningful.

csantanapr:
Related doc  [Release process problems and ideas](https://docs.google.com/document/d/1MVcIovQa4QkOvrM265TjoWnledhCxTOzA4rxYyA5y-4/edit#heading=h.5u6vd7j0omnl)

nader-ziada:
+1 to increasing the cycle period, but would want to be able to release more patch releases for various fixes or small changes between the longer release dates. This would not require customers to upgrade and at the same time help the us deliver a change to a customer that could be waiting for it. 

for example after 1.6, we can add small changes or bug fixes (if needed) in 1.6.1 and 1.6.2 and so on.

carlisia:
+1 

There is a lot of automation for the release process, but also a lot of manual verification of that automation is required. There's also some possible points of failure for those automations, which causes those to need to be run manually in that case. In other words, a release required a lot of human involvement and meticulous following of documentation. A payoff for this level of work doesn't seem to be there for a 6-week cycle.

May I ask what is a list of things needed to move on with this decision, or not? I remember hearing "write and circulate a blog post" in today's TOC meeting. Could someone please list all the things to be checked off to bring this to a conclusion?

dsimansk:
+1 & +1 to all the previous comments. My preference is to start with extending the cycle as proposed to 12 weeks. Monitor feedback and eventually consider a longer cycle, but jumping directly to something like 2 releases per year seems like a big leap. As an example from Client release currently we're able to produce in average 1-2 new features + a few bug fixes in a 6 weeks cycle. That's considerably smaller release set compared to pre-1.0 times. 

[...]

PR #1155, opened 2022-09-09
:broom: Quarterly release cadence

cardil:
Fixes knative/community#1096

## Proposed Changes

- :broom: Quarterly release cadence

### DIFF
--- a/mechanics/RELEASE-SCHEDULE.md
+++ b/mechanics/RELEASE-SCHEDULE.md
@@ -5,18 +5,21 @@ weight: 50
 type: "docs"
 ---
 
-Knative releases every 6 weeks. As much as possible, releases should be driven by automation, and repos should be ready to release at any point. It should also be possible to produce and consume nightly artifacts.
+Knative releases quarterly. As much as possible, releases should be driven by automation, and repos should be ready to release at any point. It should also be possible to produce and consume nightly artifacts.
 
-With that said, it can be useful to have a list of when future releases will happen, so this document provides a schedule for the next 6+ months of releases.
+With that said, it can be useful to have a list of when future releases will happen, so this document provides a schedule for the next 12 months of releases.
+
+## Upcoming releases
 
 | Release | Date       | EOL        | Min K8s Version | Notes                         |
 | ------- | ---------- | ---------- | --------------- | ----------------------------- |
-| 1.8     | 2022-10-04 | 2023-04-04 | 1.23            | |
-| 1.9     | 2022-11-15 | 2023-05-16 | 1.23            | |
-| 1.10    | 2023-01-10 | 2023-06-27 | 1.24            | |
+| 1.8     | 2022-10-18 | 2023-05-02 | 1.23            | |
+| 1.9     | 2023-01-24 | 2023-08-01 | 1.24            | |
+| 1.10    | 2023-04-25 | 2023-10-31 | 1.25            | |
+| 1.11    | 2023-07-25 | 2024-01-30 | TBD             | |
+| 1.12    | 2023-10-24 | 2024-04-30 | TBD            | |
 
-
-## Historical (no longer supported) releases:
+## Releases supported by community
 
 | Release | Date       | EOL        | Min K8s Version | Notes                    |
 | ------- | ---------- | ---------- | --------------- | ------------------------ |
@@ -24,6 +27,11 @@ With that said, it can be useful to have a list of when future releases will hap
 | 1.6     | 2022-07-12 | 2023-01-10 | 1.22            | |
 | 1.5     | 2022-05-31 | 2022-11-29 | 1.22            | |
 | 1.4     | 2022-04-19 | 2022-10-11 | 1.22            | |
+
+## No longer supported releases
+
+| Release | Date       | EOL        | Min K8s Version | Notes                    |
+| ------- | ---------- | ---------- | --------------- | ------------------------ |
 | 1.3     | 2022-03-08 | 2022-08-30 | 1.21            | |
 | 1.2     | 2022-01-25 | 2022-07-12 | 1.21            | |
 | 1.1     | 2021-12-14 | 2022-05-31 | 1.20            | |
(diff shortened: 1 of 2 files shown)


mechanics/RELEASE-VERSIONING-PRINCIPLES.md
```diff
@@ -46,7 +46,7 @@ Upstream reference: https://kubernetes.io/releases/patch-releases
 
 ### Knative community support window principle:
 
-We (the community) support the most recent 4 versions of Knative. The term
+We (the community) support the most recent 2 versions of Knative. The term
```
evankanderson, line 49: I think we agreed to change this to 6 months, rather than a number of releases, so that we don't need to do future gymnastics here.

mechanics/RELEASE-SCHEDULE.md
```diff
@@ -5,25 +5,33 @@ weight: 50
 type: "docs"
 ---
 
-Knative releases every 6 weeks. As much as possible, releases should be driven by automation, and repos should be ready to release at any point. It should also be possible to produce and consume nightly artifacts.
+Knative releases quarterly. As much as possible, releases should be driven by automation, and repos should be ready to release at any point. It should also be possible to produce and consume nightly artifacts.
```
evankanderson, line 8: Can we provide/record the `crontab` here as an official decision, so people don't need to go and read through #1096 and the associated doc to figure out the pattern?

mechanics/RELEASE-SCHEDULE.md
```diff
@@ -5,25 +5,33 @@ weight: 50
...
-| 1.10    | 2023-01-10 | 2023-06-27 | 1.24            | |
-
+| 1.8     | 2022-10-18 | 2023-05-02 | 1.23            | |
+| 1.9     | 2023-01-24 | 2023-08-01 | 1.24            | |
+| 1.10    | 2023-04-25 | 2023-10-31 | 1.25            | |
+| 1.11    | 2023-07-25 | 2024-01-30 | TBD             | |
+| 1.12    | 2023-10-24 | 2024-04-30 | TBD            | |
 
-## Historical (no longer supported) releases:
+## Releases supported by community
```
evankanderson, line 22: Thanks for splitting this out; the other option would be to remove the `(no longer supported)` header, so that we don't need to move things between sections as often, but this works for me as well.

mechanics/RELEASE-SCHEDULE.md
```diff
@@ -5,25 +5,33 @@ weight: 50
...
 
 | Release | Date       | EOL        | Min K8s Version | Notes                         |
 | ------- | ---------- | ---------- | --------------- | ----------------------------- |
-| 1.8     | 2022-10-04 | 2023-04-04 | 1.23            | |
-| 1.9     | 2022-11-15 | 2023-05-16 | 1.23            | |
-| 1.10    | 2023-01-10 | 2023-06-27 | 1.24            | |
-
+| 1.8     | 2022-10-18 | 2023-05-02 | 1.23            | |
+| 1.9     | 2023-01-24 | 2023-08-01 | 1.24            | |
+| 1.10    | 2023-04-25 | 2023-10-31 | 1.25            | |
```
evankanderson, line 18: I don't think we know what the min K8s version here will be. 1.25 will be the min k8s version when k8s 1.27 is released, but there's no https://github.com/kubernetes/sig-release/blob/master/releases/release-1.27/README.md file with the schedule yet.

https://github.com/kubernetes/sig-release/blob/master/releases/release-1.24/README.md was released the week of May 3, for example, which would be _after_ April 25.

mechanics/RELEASE-SCHEDULE.md
```diff
@@ -5,25 +5,33 @@ weight: 50
...
+
+## Upcoming releases
 
 | Release | Date       | EOL        | Min K8s Version | Notes                         |
 | ------- | ---------- | ---------- | --------------- | ----------------------------- |
-| 1.8     | 2022-10-04 | 2023-04-04 | 1.23            | |
-| 1.9     | 2022-11-15 | 2023-05-16 | 1.23            | |
-| 1.10    | 2023-01-10 | 2023-06-27 | 1.24            | |
-
+| 1.8     | 2022-10-18 | 2023-05-02 | 1.23            | |
```
evankanderson, line 16: Shouldn't this date be 04-25?

mechanics/RELEASE-VERSIONING-PRINCIPLES.md
cardil: That's boil down to the same outcome. 2 releases * 3 months apart.

Maybe better to keep saying about number of releases, because that's allow to calculate EOL easily, using Cron expression as well.

EOL for n release: Look for n+2 release date and add one week.

mechanics/RELEASE-SCHEDULE.md
cardil: Previously we had one week overlap. I think it's a good idea to keep it. The EOL was calculated by taking n+2 release date and adding one week to it. That way you ensure the n+2 release be already out, when n hits his EOL.

mechanics/RELEASE-SCHEDULE.md
cardil: Fixed in 5561bdd9ac744f159810d10292db49492afecea8

mechanics/RELEASE-SCHEDULE.md
cardil: Fixed in 5561bdd9ac744f159810d10292db49492afecea8

mechanics/RELEASE-VERSIONING-PRINCIPLES.md
evankanderson: If we mean the "add a week" thing to be real, we should spell it out here.

I'm okay with that, though the fact that we don't support skip-level upgrades makes that extra week less meaningful.

cardil: I added such sentence on line 15 in `mechanics/RELEASE-SCHEDULE.md`.

evankanderson (commented):
/lgtm
/approve

PR #1157, opened 2022-09-12
Update MATURITY-LEVELS.md to have concrete coverage level

salaboy:
# Changes

This is a proposal to add the coverage level for Usable and Stable components. Even if the levels are not ok, we should be explicit about how much coverage is required for both levels of maturity. 

- :gift: add link to codecov example and add 50% coverage level for Usable components.
- :gift: add link to codecov example and add 70% coverage level for Stable components.

/kind enhancement

It doesn't fix an issue because this is an update on the maturity levels that needs to be discussed with Steering. I couldn't find any other doc where these coverage levels are mentioned, as they are only configured in codecov. 

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/mechanics/MATURITY-LEVELS.md
+++ b/mechanics/MATURITY-LEVELS.md
@@ -63,7 +63,7 @@ Additional requirements beyond "initiating":
     the release artifacts. Commercial products with production usage count
     towards this target. This can be documented in something
     [like an ADOPTERS.md file](https://www.google.com/url?q=https://github.com/cncf/toc/blob/master/process/graduation_criteria.adoc&sa=D&ust=1597952611892000&usg=AFQjCNFymwghRnNGVqbD0O_01TsfEeto5w).
-- Unit test automation meeting the Knative standards with respect to coverage
+- Unit test automation meeting the Knative standards with respect to coverage for usable components (50% + using [codecov](https://app.codecov.io/gh/knative/serving))
   and flakiness / test failures.
 - Provide at least monthly progress reports at the appropriate WG(s). WGs should
   roll these reports up into their TOC reports.
@@ -93,6 +93,7 @@ Additional requirements beyond "usable":
   The API contract should be explicitly documented in the repo.
 - Identified and named set of project leads managing both development roadmap
   and overall project success.
+- Unit test automation meeting the Knative standards with respect to coverage for usable components (70% + using [codecov](https://app.codecov.io/gh/knative/serving))
 - Release schedule aligned with the core Knative release schedule.
 - At least 5 end-users consuming the release artifacts in installations of
   sufficient quality and scope.


csantanapr:
I like this additions. what's the latest status to merge this? @lance @salaboy

psschwei:
I'm ok with the levels proposed. Should we set a target date for meeting those thresholds so that components have time to meet the new standards if they aren't already (if that's even an issue)?

lance:
> I'm ok with the levels proposed. Should we set a target date for meeting those thresholds so that components have time to meet the new standards if they aren't already (if that's even an issue)?

I think that would be good. I know functions appears to have very low coverage - lower than required. However, this is (I think) primarily due to the fact that we have a suite of tests that are not getting reported in the coverage results. It will take some small amount of effort to get that worked out, but it's been a pretty low priority.

psschwei:
Does by the next release (v1.9 on Jan 24, 2023) sound like a good deadline?

lance:
> Does by the next release (v1.9 on Jan 24, 2023) sound like a good deadline?

Yes absolutely - we are currently about 0.3% away from "Usable".

lance:
> > Does by the next release (v1.9 on Jan 24, 2023) sound like a good deadline?
> 
> Yes absolutely - we are currently about 0.3% away from "Usable".

Now at 57% and rising! :)

mechanics/MATURITY-LEVELS.md
```diff
@@ -63,7 +63,7 @@ Additional requirements beyond "initiating":
     the release artifacts. Commercial products with production usage count
     towards this target. This can be documented in something
     [like an ADOPTERS.md file](https://www.google.com/url?q=https://github.com/cncf/toc/blob/master/process/graduation_criteria.adoc&sa=D&ust=1597952611892000&usg=AFQjCNFymwghRnNGVqbD0O_01TsfEeto5w).
-- Unit test automation meeting the Knative standards with respect to coverage
+- Unit test automation meeting the Knative standards with respect to coverage for usable components (50% + using [codecov](https://app.codecov.io/gh/knative/serving))
```
lance, line 66: Can we just say "Test automation" here (and below)? Our coverage in `knative/func` comes from both unit and integration tests.

evankanderson: I think integration test automation goes above and beyond.

I think "usable" should probably have unit test automation and coverage analysis. I'm okay with a separate line item saying "overall test coverage exceeds 50%" (which is not a high bar, but does require some effort in writing tests and thinking about coverage).

If we separate it into two different lines, then the line for stable could simply say "test coverage greater than 70%"

dprotaso:
@salaboy you still working on this? 

dprotaso:
Going to close this out 

cc @knative/technical-oversight-committee in case you have opinions - feel free to create a new PR

PR #1167, opened 2022-09-27
Scott stepping down from ToC

n3wscott:
# Changes

- Scott moved off ToC.


lance (approved):
Thanks for all of your many valuable contributions to the project @n3wscott! You will be missed.

/lgtm

Issue #1169, opened 2022-09-29
New Repo: `func` (move from knative-sandbox/kn-plugin-func)


salaboy:
## Repo information

Org: `knative`

Repo: `func`

Purpose (Description): 
Moving Knative Functions from `knative-sandbox`/`kn-plugin-func` to `knative`/`func`. The reasons behind this request to move are: 
- The project is now aligned with the main Knative release train
- We are working towards making the project to reach the "usable" maturity level
- We want to announce this new branding `func` during KubeCon NA / KnativeCon NA 2022

Sponsoring WG: Knative Functions Working Group

## Actions to fulfill

This area is used to track the [repo creation process](https://github.com/knative/community/blob/main/mechanics/CREATING-A-SANDBOX-REPO.md).
The _requestor_ and _sponsoring WG lead_ should perform the steps listed below and cross out the checkmarks when done.
The TOC is involved only in the **TOC Gate** steps.

- [x] Add this issue to the [TOC project board](https://github.com/orgs/knative/projects/9) for review. You are responsible for moving your entry on the board to "Needs Discussion" or "In Progress" as you move forward in this checklist.

_You may not be able to use the Projects quick menu on this page. In that case, go to the project board and use the **Add cards** interface._

- [x] Send a PR adding entries for this repo in [`/peribolos/knative-sandbox.yaml`](https://github.com/knative/community/blob/main/peribolos/knative-sandbox.yaml). Please mind the alphabetical order when adding to a list.
  - [x] Add the repository and a description.
  - [x] Grant `Knative Admin` the `admin` privilege.
  ~- [ ] Grant the sponsoring WG the `write` privilege.~

Note that for repos moving from knative-sandbox, it is necessary to move these entries from knative-sandbox.yaml to knative.yaml

**TOC Gate**: _Once the TOC has approved the above, it will merge and Peribolos will create an empty repository._

- [x] Ask Steering to add the repo to EasyCLA in [Slack](https://knative.slack.com/archives/CU9DGL4FM).

- [x] (golang) Send a PR to add aliases for `knative.dev/$REPONAME` import paths ([sample](https://github.com/knative/docs/pull/4160)).

~~- [ ] Have a lead from the sponsoring WG bootstrap the Git repository by pushing an~~
  ~appropriate "template" repository ([basic](https://github.com/knative-sandbox/wg-repository),~
  ~[sample-controller](https://github.com/knative-sandbox/sample-controller),~
[...]

PR #1170, opened 2022-09-29
chore: add entries for knative/func

lance:
This commit moves knative-sandbox/kn-plugin-func and the team members over to knative/func.

See: https://github.com/knative/community/issues/1169

### DIFF
--- a/peribolos/knative-sandbox-OWNERS_ALIASES
+++ b/peribolos/knative-sandbox-OWNERS_ALIASES
@@ -114,16 +114,13 @@ aliases:
   - nainaz
   - salaboy
   - zroubalik
-  func-writers:
+  functastic-writers:
   - lance
   - lkingland
   - matejvasek
   - nainaz
   - salaboy
   - zroubalik
-  functions-wg-leads:
-  - lance
-  - salaboy
   homebrew-kn-plugins-approvers:
   - dsimansk
   - maximilien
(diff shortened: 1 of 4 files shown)


peribolos/knative-sandbox-OWNERS_ALIASES
```diff
@@ -114,16 +114,13 @@ aliases:
   - nainaz
   - salaboy
   - zroubalik
-  func-writers:
+  functastic-writers:
```
psschwei, line 117: For a lot of the other groups, we use the same reader/write/leads groups across both the knative/sandbox lists (so for example, we have serving-writers in both peribolos files, with the same names in each). Then you can also set up separate groups for sandbox repos if you want to allow someone to approve the sandbox but not the core repo ([example](https://github.com/knative-sandbox/container-freezer/blob/main/OWNERS#L4-L5)). Wonder if that might work here too (just thinking out loud, so feel free to disregard)?

salaboy:
does this needs to stay on hold until the repo is finally moved?

lance:
> does this needs to stay on hold until the repo is finally moved?

Yes I think so, unless we want to do it in two steps - add everything for `knative/func` in this PR, and remove from `knative-sandbox/kn-plugin-func` in a second PR.

evankanderson:
/lgtm

Remove the hold when we're ready to pull the trigger.

evankanderson (approved):
/hold cancel

PR #1166, opened 2022-09-27
make steering world postable. Fix description.

vaikas:
Signed-off-by: Ville Aikas <vaikas@chainguard.dev>

# Changes

🐛 Emails sent to steering are bouncing. This makes the list publicly postable.
@evankanderson 
@csantanapr 

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

PR #1177, opened 2022-10-04
Allow `elections@knative.team` to receive emails from the world

upodroid:
Applying https://github.com/knative/community/pull/1166 to elections group as well

/cc @lance


lance (commented):
/lgtm
/approve

PR #1184, opened 2022-10-21
Removes inactive member to fix func owners issue

psschwei:
Signed-off-by: Paul S. Schweigert <paul@paulschweigert.com>

# Changes

As discussed in [Slack](https://knative.slack.com/archives/CCSNR4FCH/p1666360097149169), there's an inactive member who never accepted the invite to the org, which is causing problems with the owners files in the func repo. As the person has had no github activity in over a year (and also no knative activity), this PR drops that person from the org list.

cc @lance @upodroid


upodroid (approved):
LGTM

PR #1186, opened 2022-10-25
Add `nainaz` to voters for 2022 SC Election

upodroid:
## Voting Exception Request

@nainaz messaged me and asked to be added an eligible voter.


geekygirldawn:
We actually have a process for this. @nainaz - can you please go to https://elections.knative.dev/app/elections/2022-SC and fill out the exception request form?

We also handle the approvals privately between the election offices and the SC to avoid publicly making anyone feel bad if their exception request is declined.

upodroid:
Sorry

/close

PR #1105, opened 2022-06-30
Update Trademark composition

evankanderson:
/assign @thisisnotapril

(For Mark; Simon's assignment was already done in the `knative/peribolos.yaml` file for ACLs, but not at the top-level charter doc.)

PR #1189, opened 2022-10-31
Updating peribolos with TM committee changes

psschwei:
Signed-off-by: Paul S. Schweigert <paul@paulschweigert.com>

Mark was added to TM in PR #1105, this adds that change to peribolos.

This also fixes a blocker for functions: https://github.com/knative/func/pull/1342#issuecomment-1296432869 

cc @mchmarny


lance (commented):
/lgtm

PR #1198, opened 2022-11-03
Add lionel as eventing writer

lionelvillard:
https://github.com/knative/community/pull/1192 follow-up.

/cc @pierDipi


pierDipi (commented):
Powerless LGTM.

/lgtm

PR #1202, opened 2022-11-07
Voter exceptions for 2022 SC Election

upodroid:
Adding voters who requested exceptions to the 2022 SC Election

I'll keep adding people whose requests have been accepted. I will merge the PR after the deadline has passed (28th of Nov).

/hold


geekygirldawn:
/hold cancel
/lgtm

We can merge this now. We don't need to collect all of the exceptions - we usually do each one as a PR.

PR #1204, opened 2022-11-07
Add self nomination for 2022 SC election

aslom:
# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/elections/2022-SC/candidate-aslom.md
+++ b/elections/2022-SC/candidate-aslom.md
@@ -1,0 +1,10 @@
+-------------------------------------------------------------
+name: Aleksander Slominski
+ID: aslom
+info:
+  - affiliation: IBM
+-------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


geekygirldawn:
/hold
Need to verify eligibility and make sure the format is what Elekto expects.

geekygirldawn:
I've confirmed that @aslom is eligible to run:
* Is an eligible voter in this election (confirmed in voters.yaml).
* Holds at least a project member role (is an project member and reviewer for Eventing)

@jberkus the format looks good to me, but given the disastrous results of the last merge, can you confirm that the format is what Elekto expects? If so, I think you can remove the hold and merge.

jberkus:
Format is fine.

/remove-hold
/lgtm

PR #1211, opened 2022-11-15
Fixing pierDipi's eligibility in voters.yaml

geekygirldawn:
Changing case of @pierDipi's gh handle in voters.yaml to see if that solves the issue with eligibility in Elekto


upodroid (approved):
LGTM

PR #1215, opened 2022-11-21
voters.yaml update for AngeloDanducci

geekygirldawn:
changed the case for AngeloDanducci's voters.yaml gh id as a workaround to the Elekto case bug to allow voter eligibility in Elekto


upodroid (approved):
LGTM

PR #1216, opened 2022-11-22
Fixed GH login case for several users in voters.yaml

geekygirldawn:
Fixed the remainder of the eligible voters that didn't match the case of their GH login with the voters.yaml file to get around the bug in Elekto.


upodroid (approved):
LGTM

PR #1222, opened 2022-12-07
Release 1.5 is no longer supported

dprotaso:
/assign @psschwei


psschwei:
Did we extend the support window for v1.5 when we moved to the quarterly release cycle? I vaguely recall that being discussed at one point, so just wanted to double check.

psschwei:
Actually, I think I was thinking of this line from the proposal: `For a quarterly release cycle, this would involve supporting the 2 most recent releases (6 months).` so the changes don't apply to v1.5

psschwei (commented):
/lgtm
/approve

PR #1243, opened 2023-01-11
Add dave to privileged infra group

upodroid:
/cc @dprotaso 

This should give you sufficient privileges to troubleshoot.


upodroid:
/override pull-knativeteam-groups-test

Don't really need the test to pass.

PR #1244, opened 2023-01-12
Drop in-active members

dprotaso:
Queried a list of users who haven't interacted with any of the repos in knative/knative-sandbox in 2022 using https://www.gharchive.org/

BigQuery was 

```SQL
  SELECT DISTINCT actor.login, COUNT(*)
  FROM `githubarchive.year.2022`
  WHERE org.login IN ('knative', 'knative-sandbox')
  GROUP BY actor.login
```

If I made a mistake it should be easy to add again

(for the future) Made a terrible script to do the diff
https://gist.github.com/dprotaso/e412da7b1652c3705a42c497891eadfa


dprotaso:
peribolos won't allow doing a big delete like this without an explicit override
> cannot delete 382 memberships or 0.753 of knative (exceeds limit of 0.250)

psschwei:
+1

Looks like this would be a drop of around 80% of the current members of the knative org.

psschwei (commented):
/lgtm

PR #1271, opened 2023-02-13
Fix permissions for .artifacthub repo

dsimansk:
@psschwei PTAL. I've just noticed that `admins` should definitely `admin` the repo.


psschwei (commented):
/lgtm
/approve

PR #1281, opened 2023-02-15
drop kind/good-first-issue label

dprotaso:
We should be using 'good first issue' instead. As that's what the prow `HELP` plugin uses

See prow commands
https://prow.knative.dev/command-help

### DIFF
--- a/label_sync/labels.yaml
+++ b/label_sync/labels.yaml
@@ -29,12 +29,6 @@ default:
       target: issues
       prowPlugin: label
       addedBy: humans
-    - color: 7057ff
-      description: Denotes an issue ready for a new contributor.
-      name: kind/good-first-issue
-      target: issues
-      prowPlugin: label
-      addedBy: anyone
     - color: e11d21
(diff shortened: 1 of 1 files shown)


dprotaso:
/assign @kvmware @upodroid 

I'm assuming we don't need to explicitly add the `good first issue` and `help wanted` labels correct?

kvmware:
I personally would like to remove all the `kind/` from all the `kind/*` labels, but @upodroid has different opinion last I checked.
I need to remove https://github.com/knative/community/blob/ed454ebe08439bd8c0eb9309a6051eaa04d1c479/label_sync/labels.yaml#L99-L110 as well.
As far as `good first issue` etc the label syncer specified labels are synced everywhere but does not remove manually added labels. I would say it probably is good to have labels that are org wide to be specified but if there are already there then no harm.

upodroid (approved):
LGTM

We should use the more widely used label for identifying good first issues.

PR #1282, opened 2023-02-15
use OWNER aliases vs. specific people

dprotaso:


label_sync/OWNERS
```diff
@@ -1,10 +1,8 @@
 # The OWNERS file is used by prow to automatically merge approved PRs.
 
 approvers:
-- steering-committee
```
dprotaso, line 4: already inherited from top level OWNER files

groups/OWNERS
```diff
@@ -8,9 +8,9 @@ options:
   no_parent_owners: true
 
 approvers:
-- dprotaso
-- upodroid
-- kvmware
```
dprotaso, line 13: adding TOC here so it's not just me :)

upodroid (approved):
LGTM

PR #1287, opened 2023-02-21
Allow release leads to submit jobs to prow via kubectl

upodroid:
/cc @kvmware


kvmware:
/lgtm
THANKS!!!

kvmware:
Can we add the release leads group there too?

PR #1292, opened 2023-02-23
update review schedule

dprotaso:
No quorum at today's meeting so pushing back the schedule.


dprotaso:
/assign @zroubalik @dsimansk 

cc @evankanderson for the Security WG update next week (hope you're feeling better 🤒 )

evankanderson (commented):
/lgtm

I think we might want to reconsider the frequency as we've consolidated working groups; it feels like slightly more than quarterly was a good frequency for updates.

(I'm still happy to do security WG next week)

dprotaso:
> I think we might want to reconsider the frequency as we've consolidated working groups; it feels like slightly more than quarterly was a good frequency for updates.

Yeah good point

PR #1312, opened 2023-03-13
Add davidhadas to security group mailing list

davidhadas:
add david.hadas@gmail.com

PR #1313, opened 2023-03-14
Drop paul, julz from security mailing list group

psschwei:
# Changes

Follow up to #1312, dropping myself from the security mailing list in favor of David. Also dropping julz, as he's no longer active here.

/assign @evankanderson


evankanderson (commented):
/lgtm
/approve

psschwei:
looks like there's a bit of confusion with one of the actions, let's see if this fixes it
/override "Verify Deps and Codegen (1.17.x, ubuntu-latest)"

PR #1321, opened 2023-04-06
Grant @jeefy GCP admin access

upodroid:
/cc @jeefy

This group has all these privileges on the knative.dev org

https://github.com/knative/test-infra/blob/main/infra/gcp/iam.tf


puerco:
Thanks @upodroid !
/lgtm

PR #1345, opened 2023-05-10
Create candidate-davidhadas.md

davidhadas:
Create candidate-davidhadas.md


aliok:
I've confirmed that @davidhadas is [eligible as a candidate](https://github.com/knative/community/blob/main/mechanics/TOC.md#candidate-eligibility) for the TOC election as an approver with more than 3 months tenure.

I also looked at the format of the candidate file, and it looks correct.

cc @geekygirldawn

PR #1360, opened 2023-05-24
HoldTillRename: Rename knative-sandbox to knative-extensions

davidhadas:
This is an auotmated PR to replace  strings to . Do not merge until Rename knative-sandbox org is renamed to knative-extensions


davidhadas:
replaced with knative-extension

PR #1370, opened 2023-06-01
2023 TOC election results

aliok:
As suggested in https://github.com/knative/community/tree/main/elections#process, publishing 2023 TOC election results, so that it shows up in Elekto properly.

Text copied from the SC announment.


geekygirldawn (approved):
/lgtm

Thanks to everyone for all of their help with the election 🎉

Issue #1369, opened 2023-06-01
Onboard new TOC member Krsna


aliok:
The following tasks need to be done 

- Add @kvmware as new member
- Remove @evankanderson  

- [ ] Add new member in [TECH-OVERSIGHT-COMMITTEE.md](https://github.com/knative/community/blob/main/TECH-OVERSIGHT-COMMITTEE.md) https://github.com/knative/community/pull/1368
- [ ] [peribolos](https://github.com/knative/community/tree/main/peribolos) for both knative and knative-sandbox https://github.com/knative/community/pull/1371
- [ ] groups for mailing lists https://github.com/knative/community/pull/1372
  - https://github.com/knative/community/blob/main/groups/committee-oversight/groups.yaml
  - https://github.com/knative/community/blob/main/groups/groups.yaml#L60
- [ ] [CNCF Project maintainers](https://github.com/cncf/foundation/blob/main/project-maintainers.csv)
- [ ] CNCF Knative Project JIRA access (via [cncf service desk](https://cncfservicedesk.atlassian.net/servicedesk/customer/portal/1) Jira ticket)
- [ ] CNCF Knative maintainers group list (via [cncf service desk](https://cncfservicedesk.atlassian.net/servicedesk/customer/portal/1) Jira ticket)

psschwei:
In case it helps, I've got a doc in the TOC folder on the shared drive with all the things I had to do when I onboarded

kvmware:
The CNCF parts I think need an existing member or steering. If there is anything else I can do please let me know.

aliok:
Krsna is already in the TOC and  I am guessing after a year, this has been completed.

/close

Issue #169, opened 2020-06-25
Rename knative-sandbox github org


bsnchan:
Historical context: #69

Based off of the Functions WG retro, the `knative-sandbox` name is confusing as it contains stable code. We currently have 3 use cases:

* experiments/incubation: things we hope to grow into core, extensions, or abandon
* extensions that implement an interface
* "core" interfaces with multiple implementations

Some ideas from the conversation:
* knative-contrib
* knative-extensions
* knative-net
* knative-experimental
* knative-incubation

Please add your ideas/proposals in this issue

rhuss:
I'm plus 👍  for using another name as sandbox as the name is really associated as being kind of a playground.

If there is supposed to be only one single organisation, then I think something generic like `knative-contrib` or `knative-extensions` makes much sense.

However, we can also have multiple organisations reflecting the different purpose as mentioned above:

* `knative-sandbox` for experiments and incubation to other orga except knative proper.
* `knative-extensions` for extensions/plugin that can hook into a Knative installation. Examles for this would be the network repos, event sources or client plugins
* `knative-contrib` everything which adds to the Knative ecosystems, but which goes beyond an POC or R&D exeperiment.

I could imagine, that `knative-sandbox` _could_ be the incubation stage for `knative-contrib` & `knative-extensions` but I wouldn't enforce that (i.e. it should be possible to create are repo directly in `knative-extensions` or `knative-contrib` without going through `knative-sandbox`).

If 4 orgas are too much of an overhead I could imagine to skip `knative-sandbox` as experiments can happen elsewhere, too. Also `knative-extensions` could fold with `knative-contrib` if this is too much.

So, I can imagine kind of three setup for orgas in addition to `knative`:

* `knative-sandbox` : PoC and R&D kind of projects, not supposed to be production ready
* `knative-extensions`:  Implementations of `knative` core hooks like network, event sources, even channel implementations, client plugins
* `knative-contrib` : Anything else adding to the Knative ecosystem like addition client tooling, integration into platforms like serverless.com etc.

**OR:**

* `knative-extensions`:  Implementations of `knative` core hooks like network, event sources, even channel implementations, client plugins
* `knative-contrib` : Anything else adding to the Knative ecosystem like addition client tooling, integration into platforms like serverless.com etc.

and experimental projects life somewhere else (e.g. in personal accounts)

**OR:**

* `knative-contrib` : Implementations of `knative` core hooks as described above + anything else.

markusthoemmes:
[...]

PR #1371, opened 2023-06-01
Add Krsna to TOC and infra WG leads as admins

krsna-m:
part of https://github.com/knative/community/issues/1369

Note adding Mahamad and Krsna as org admins to handle knative-sandbox rename (https://github.com/knative/community/issues/169)

### DIFF
--- a/peribolos/knative-sandbox.yaml
+++ b/peribolos/knative-sandbox.yaml
@@ -14,7 +14,6 @@ orgs:
     - csantanapr
     - dprotaso
     - dsimansk
-    - evankanderson
     - itsmurugappan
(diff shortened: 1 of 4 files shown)


peribolos/knative-sandbox.yaml
```diff
@@ -707,11 +707,11 @@ orgs:
           Technical Oversight Committee:
             description: Members of the Knative Technical Oversight Committee (TOC)
             maintainers:
-            - evankanderson
```
psschwei, line 710: I think you also need to move Evan to the members list (and remove him from the admins one) in both knative and sandbox

PR #1382, opened 2023-06-15
updated my email address with my personal one, since I'm leaving VMwa…

geekygirldawn:
Updated my email address for the elections email alias with my personal one, since I'm leaving VMware as of July 3.


dprotaso:

PR #1385, opened 2023-06-17
Hand off VMware trademark seat to Sameer

evankanderson:
Since Trademark seats are based on company affiliation, hand off the VMware seat to Sameer now that I've left VMware.

### DIFF
--- a/peribolos/knative-OWNERS_ALIASES
+++ b/peribolos/knative-OWNERS_ALIASES
@@ -156,7 +156,7 @@ aliases:
   - psschwei
   - zroubalik
   trademark-committee:
-  - evankanderson
+  - xtreme-sameer-vohra
   - mchmarny
   - smoser-ibm
   ux-wg-leads:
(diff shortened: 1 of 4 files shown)


peribolos/knative-OWNERS_ALIASES
```diff
@@ -156,7 +156,7 @@ aliases:
   - psschwei
   - zroubalik
   trademark-committee:
-  - evankanderson
+  - xtreme-sameer-vohra
```
dprotaso, line 159: run codegen - sees like user names are sorted here

dprotaso:
@evankanderson can you re-run codegen for this PR

kvmware:
@dprotaso Ran it for him see my commit. Hope @evankanderson is ok with that :pray:

evankanderson:
Thanks, I think I was doing this on my windows machine where the autogen stuff is not always so happy.

/approve

dprotaso:
cc @knative/steering-committee for approval

PR #1387, opened 2023-06-27
Update leads for 1.11

skonto:
# Changes

- Updates leads As described in https://github.com/knative/release/blob/main/PROCEDURES.md#permissions-for-release-leads

### DIFF
--- a/peribolos/knative-OWNERS_ALIASES
+++ b/peribolos/knative-OWNERS_ALIASES
@@ -56,6 +56,8 @@ aliases:
   - lance
   - salaboy
   knative-admin:
+  - Vishal-Chdhry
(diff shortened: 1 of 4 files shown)


peribolos/knative-OWNERS_ALIASES
```diff
@@ -56,6 +56,8 @@ aliases:
   - lance
   - salaboy
   knative-admin:
+  - Vishal-Chdhry
+  - craydr
```
pierDipi, line 60: That's mistyped cc @creydr

creydr: opened #1389

Issue #1383, opened 2023-06-15
PROCESS CHANGE: WG leads should be in CNCF's maintainers.csv


aliok:
During https://github.com/knative/community/issues/945 I see TOC and SC members are added to the list in https://github.com/cncf/foundation/blob/main/project-maintainers.csv.

However, we should have more people there.

For example, Istio has different groups (https://github.com/cncf/foundation/blob/main/project-maintainers.csv#L1179):
- Istio: Steering Committee
- Istio: Technical Oversight Committee
- Istio: Working Group Leads
- Istio: Maintainers

We can go with the same structure.

## Expected benefits

- Rewarding the contributors by listing their name there
- Being able to subscribe to maintainers-only things like some newsletters, etc.
- Getting rid of the necessity to prove that I am actually a maintainer in Knative project.

## Expected costs
- More people are listed there - shouldn't cause any issue
- Scripting change

**UPDATE**: see this comment: https://github.com/knative/community/issues/1383#issuecomment-1865809475 . We'll add WG leads and not all approvers.

lance:
> * @lance to lead the discussion for those criteria

I've sent a PR with some language for removal - voluntary and not - in the ROLES.md document here: https://github.com/knative/community/pull/1390. Feedback is welcome - there are a few questions and consider it a WIP until we come to agreement.

aliok:
Update, in SC meeting, we had a vote and agreement for having WG leads in maintainers.csv file as well: https://docs.google.com/document/d/16Rpd2nhmLWrkFJUfpJ31Ox2PpultJ-bYIhqaiWQff_I/edit#bookmark=id.8jsaovwwr9fw

Not all approvers, but WG leads (in addition to SC+TOC members)

Updating the ticket description above.

aliok:
PR https://github.com/cncf/foundation/pull/720

dprotaso:
Going to close this out - I send cmierly@LF an updated list

PR #1326, opened 2023-04-11
Ashleigh stepping down as Docs WG lead

abrennan89:
# Changes

Removing myself as WG lead

/kind removal

PR #1390, opened 2023-07-03
[PROCESS CHANGE]: add a simple inactivity measure and process for removal

lance:
The ROLES.md document was recently updated to more accurately reflect how the Knative organization should work with regards to a "contribution ladder" and methods of moving up and down that latter based on project contributions. What was not addressed in those changes was how to handle stepping away from the project or being removed for some reason.

This change is a first pass at addressing that omission. The text is taken directly from the CNCF contribution ladder template located here:

https://github.com/cncf/project-template/blob/main/CONTRIBUTOR_LADDER.md#inactivity

What still remains unaddressed is what, if anything happens in peribolos. I think that should probably be stated somewhere in this document or another.

Clearly, when users step away from the project or WG leads change, they should be removed from teams that have elevated priviledges (writers, reviewers, admin, etc). To date, this just mostly means that they end up remaining members of the knative and knative-sandbox orgs, but no longer have write priveleges for specific repos or administrative capabilities.

It's not clear to me, however, what is the best way to deal with people stepping away from the project entirely or being removed for some cause (e.g. violation of the code of conduct). Users in the knative and knative-sandbox orgs in Peribolos already have restricted, read-only rights. Does it make sense to create an `emeritus` group to explicitly move these users into a recognized team that still has restricted, read-only rights but is recognized explicitly as no-longer contributing?

Related: https://github.com/knative/community/issues/1383

/kind documentation
/hold as WIP


ROLES.md
```diff
@@ -276,6 +276,28 @@ The following apply to the area / component for which one would be a lead.
...
+## Involuntary Removal or Demotion
+
+Involuntary removal/demotion of a contributor happens when responsibilities and requirements aren't being met. This may include repeated patterns of inactivity, extended period of inactivity, a period of failing to meet the requirements of your role, and/or a violation of the Code of Conduct. This process is important because it protects the community and its deliverables while also opens up opportunities for new contributors to step in.
+
+Involuntary removal or demotion is handled through a vote by a majority of the current Steering Committee members.
+
+## Stepping Down/Emeritus Process
+If and when contributors' commitment levels change, contributors can consider stepping down (moving down the contributor ladder) vs moving to emeritus status (completely stepping away from the project).
+
+Contact the Maintainers about changing to Emeritus status, or reducing your contributor level.
```
aliok, line 299:
```suggestion
Contact the Steering Committee about changing to Emeritus status, or reducing your contributor level.
```

aliok:
(non-binding comment)

> This change is a first pass at addressing that omission. 

LGTM!

> What still remains unaddressed is what, if anything happens in peribolos. I think that should probably be stated somewhere in this document or another.
> ...
> Does it make sense to create an emeritus group to explicitly move these users into a recognized team that still has restricted, read-only rights but is recognized explicitly as no-longer contributing?

It is reasonable. @puerco mentioned in the call today that some SIG (can't remember which one) is doing it this way and I think it makes sense.

Thanks for driving this @lance

lance:
I've added some verbiage around Peribolos changes. We should create that group before landing this PR.

geekygirldawn:
Thanks for addressing this, and this looks like a good way of handling it.

aliok:
I just saw [this](https://github.com/cncf/toc/blob/main/process/graduation-proposal-template.md#-explicitly-define-the-criteria-process-and-offboarding-or-emeritus-conditions-for-project-maintainers-or-those-who-may-interact-with-the-cncf-on-behalf-of-the-project-the-list-of-maintainers-should-be-preferably-be-stored-in-a-maintainersmd-file-and-audited-at-a-minimum-of-an-annual-cadence) in the graduation criteria:

> Explicitly define the criteria, process and offboarding or emeritus conditions for project maintainers; or those who may interact with the CNCF on behalf of the project. The list of maintainers should be preferably be stored in a MAINTAINERS.md file and audited at a minimum of an annual cadence.

This PR change addresses some of that requirement, which is great. 

We also need an audit too. Let's not worry about that in this PR though.

lance:
I have added an Emeritus group to Peribolos, and included for now @abrennan89 and @jcrossley3 since they have both definitely stepped away from the project. I think actually moving other folks who have stopped contributing should happen in a separate PR. This is just being done for illustrative purposes.

aliok:
> I have added an Emeritus group to Peribolos

ping @knative/productivity-leads

upodroid:
The best course of action is removing someone from GitHub teams. Privileges are granted via GitHub Teams and prow checks GitHub Teams when considering /approve commands.

Being an org member is effectively having 0 permissions given we don't have private repos.

`emeritus_approvers` aren't that helpful as there should be no GitHub usernames in any OWNERS files.

ROLES.md
```diff
@@ -276,6 +276,30 @@ The following apply to the area / component for which one would be a lead.
...
+## Involuntary Removal or Demotion
+
+Involuntary removal/demotion of a contributor happens when responsibilities and requirements aren't being met. This may include repeated patterns of inactivity, extended period of inactivity, a period of failing to meet the requirements of your role, and/or a violation of the Code of Conduct. This process is important because it protects the community and its deliverables while also opens up opportunities for new contributors to step in.
+
+Involuntary removal or demotion is handled through a vote by a majority of the current Steering Committee members.
+
+## Stepping Down/Emeritus Process
+If and when contributors' commitment levels change, contributors can consider stepping down (moving down the contributor ladder) vs moving to emeritus status (completely stepping away from the project).
+
+Contact the Steering Committee about changing to Emeritus status, or reducing your contributor level.
```
psschwei, line 299: FWIW, in the past, stepping down has been handled by the person in question opening a PR (#1326 for example). 

Also, currently promoting a person to reviewer/approver is handled by the WG leads, while promoting someone to lead is done by the TOC. Would we want to have promotion and demotion handled by the same group of folks?

ROLES.md
```diff
@@ -276,6 +276,30 @@ The following apply to the area / component for which one would be a lead.
...
 
+
+## Inactivity
+It is important for contributors to be and stay active to set an example and show commitment to the project. Inactivity is harmful to the project as it may lead to unexpected delays, contributor attrition, and a lost of trust in the project.
+
+* Inactivity is measured by:
+    * Periods of no contributions for longer than 12 months
+    * Periods of no communication for longer than 12 months
+* Consequences of being inactive include:
+    * Involuntary removal or demotion
```
psschwei, line 287: I'm a bit torn on the idea of demotion... given that a member is defined as "an established Contributor who regularly participates in the project", I could see an argument for just straight dropping folks who haven't contributed in a year.

kvmware: I like the idea of just removing inactive folks instead of demoting as well. Demoting, for me at least, would be a bit offensive to me. I would rather just be silently removed.

lance:
@knative/productivity-leads et. al. I've made some changes to the text, removing the emeritus team and simplifying the process of voluntary and involuntary removal. PTAL 🙏

ROLES.md
```diff
@@ -276,6 +276,26 @@ The following apply to the area / component for which one would be a lead.
...
+    * Periods of no contributions for longer than 12 months
+    * Periods of no communication for longer than 12 months
+* Consequences of being inactive include:
+    * Involuntary Removal or demotion
+
+## Involuntary Removal or Demotion
+
+Removal or demotion of a contributor happens when responsibilities and requirements aren't being met. This may include repeated patterns of inactivity, extended period of inactivity, a period of failing to meet the requirements of your role, and/or a violation of the Code of Conduct. This process is important because it protects the community and its deliverables while also opens up opportunities for new contributors to step in.
+
+Involuntary removal or demotion is handled through a vote by a majority of the current Steering Committee members.
```
psschwei, line 297: Do we want to be a bit more stringent here, maybe requiring 75% in favor of (i.e. a 4-1 vote)? I'd assume removing someone should be fairly unanimous and that if there's two folks opposed we could potentially end up in a tricky spot...

lance: Pinging @knative/steering-committee again... 

I can see that a 75% threshold would be good in contentious scenarios, but I imagine most of the time we'll just be purging the ranks of folks who have left the project for one reason or another without voluntarily removing themselves.

rhuss: +1 for a 75% consensus, I also don't see it critical, and most of the time, people probably just "forgot" to remove themselves. Maybe we could add a note that the removed person should be notified about this with a short explanation so that such an involuntary removal does not come with a surprise

lance: @rhuss @psschwei I have modified the doc to indicate an 80% vote in favor of removal (4-1) ptal

psschwei:
+1 from me (though I'm not on steering, so don't think my vote counts :smile: )

lance:
> +1 from me (though I'm not on steering, so don't think my vote counts 😄 )

@knative/steering-committee ptal

ROLES.md
```diff
@@ -276,6 +276,26 @@ The following apply to the area / component for which one would be a lead.
...
+
+* Inactivity is measured by:
+    * Periods of no contributions for longer than 12 months
+    * Periods of no communication for longer than 12 months
+* Consequences of being inactive include:
+    * Involuntary Removal or demotion
+
+## Involuntary Removal or Demotion
+
+Removal or demotion of a contributor happens when responsibilities and requirements aren't being met. This may include repeated patterns of inactivity, extended period of inactivity, a period of failing to meet the requirements of your role, and/or a violation of the Code of Conduct. This process is important because it protects the community and its deliverables while also opens up opportunities for new contributors to step in.
```
puerco, line 295: I wonder if we should split the language concerning involuntary removals due to inactivity and disciplinary reasons (eg CoC violations). My feeling is that it sets a bad precedent where er are bucketing inactive contributors together with other kinds of removals which are punitive in nature.

aliok, line 295:
```suggestion
Removal or demotion of a contributor happens when responsibilities and requirements aren't being met. This may include repeated patterns of inactivity, extended period of inactivity, a period of failing to meet the requirements of your role. This process is important because it protects the community and its deliverables while also opens up opportunities for new contributors to step in.

Similarly, a disciplinary action such as a violation of the Code of Conduct, will result in a removal or demotion of a contributor.
```

Changed some text based on @puerco 's comment https://github.com/knative/community/pull/1390/files#r1293698887

aliok: cc @lance @puerco

csantanapr:
I took @aliok recommendation and committed

jberkus:
Belatedly, +1 from me.

PR #1397, opened 2023-07-12
adding dan-j as knative-sandbox member

dan-j:
# Changes

- Adding myself, @dan-j, as a member of the knative-sandbox org.

# Reasoning

I have contributed NATS JetStream support to the eventing-natss repo, and would like to continue maintenance in the long term.


dprotaso:
/ok-to-test


You'll need to run `./hack/update-codegen.sh` script

dprotaso:
Actually maybe not

PR #1398, opened 2023-07-13
chore: update the link in adopters.md to point to a new issue

lance:
Instead of piling on the old issue that was really just asking for adopters to add a comment to the issue, we now have a specific issue form for adding adopters. This commit just updates the link in ADOPTERS.md to point to the new issue template.


psschwei (commented):
/lgtm

PR #1411, opened 2023-08-01
HoldTillRename: Rename knative-sandbox to knative-extensions

davidhadas:
This is an auotmated PR to replace  strings to . Do not merge until Rename knative-sandbox org is renamed to knative-extensions


peribolos/knative-sandbox.yaml
```diff
@@ -1,5 +1,5 @@
 orgs:
```
upodroid, line 1: Can you rename this file?

peribolos/knative-sandbox-OWNERS_ALIASES
```diff
@@ -1,5 +1,5 @@
 # This file is auto-generated from peribolos.
```
upodroid, line 1: Can you rename this file as well?

upodroid (changes requested):

PR #1426, opened 2023-09-01
Drop the go policy consistency clause

dprotaso:
The version in the go.mod file is the min go version required to
build the repository. We should always be using the new version
of the tool chain

Dropping this clause is a reflection of what we are doing today


dprotaso:
@knative/technical-oversight-committee - if someone thinks this needs a formal discussion at the TOC meeting please leave a comment expressing that.

Otherwise doing this async with +1's should suffice until we have quorum.

psschwei (commented):
+1

kvmware:
fine by me

psschwei:
That's 3/4
/lgtm

PR #1439, opened 2023-09-20
Update new repo template

dprotaso:
- No need to configure EasyCLA per repo
- Knobots shouldn't be optional - this updates the community files


psschwei (commented):
/lgtm

PR #1476, opened 2023-12-07
Drop emeritus leads from various group

psschwei:
Noticed some of these files were a bit out of date, so took a best guess at who should be dropped. 

Wasn't sure if the admins for automation@knative.team are the TOC, so didn't update that one.

cc @knative/technical-oversight-committee @knative/productivity-leads

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -28,6 +28,8 @@ groups:
       - navidshaikh@knative.team
       - nshaikh@redhat.com
       - pierdipi@knative.team
+      - rhuss@knative.team
+      - rhuss@redhat.com
       - shou@us.ibm.com
 
   - email-id: calendar-maintainers@knative.team
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -15,48 +15,22 @@ groups:
       - evan.k.anderson@gmail.com
       - evana@vmware.com
       - evankanderson@knative.team
-      - julz.friedman@uk.ibm.com
-      - julz@knative.team
-      - markusthoemmes@knative.team
-      - markusthoemmes@me.com
-      - me.markusthoemmes@googlemail.com
-      - rhuss@knative.team
-      - rhuss@redhat.com
-      - tcnghia@gmail.com
       - paul@paulschweigert.com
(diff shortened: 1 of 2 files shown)


groups/groups.yaml
```diff
@@ -15,48 +15,20 @@ groups:
       - evan.k.anderson@gmail.com
       - evana@vmware.com
       - evankanderson@knative.team
-      - julz.friedman@uk.ibm.com
-      - julz@knative.team
-      - markusthoemmes@knative.team
-      - markusthoemmes@me.com
-      - me.markusthoemmes@googlemail.com
-      - rhuss@knative.team
-      - rhuss@redhat.com
```
cardil, line 24: @rhuss is an active Knative contributor and a WG lead. Maybe he should be moved from owners to members?!

aliok: I see other WG leads are in `members`, so, that makes sense.

aliok: Or let me ask... What's `owners` and what's `members` here?

Owners --> TOC members?
Members --> TOC members + WG leads?

aliok:
FYI, this might be a blocker for https://github.com/knative/community/issues/1383

groups/groups.yaml
upodroid: Having owners doesn't really matter because it is a mailing list that the public can't join. owners can be moved to members and left empty entirely.

cardil: Either way, we can't just remove Roland here, as he isn't listed in members. That'll effectively remove him from the list.

So either restore him here, or move everyone to members as suggested.

psschwei: whoops! re-added roland, thanks for catching that

aliok: @cardil @upodroid Are we good with this one? Sorry for the pressure, but I want to go ahead and work on https://github.com/knative/community/issues/1383

Cali0707:
Looks good to me, are there any changes still needed or can we merge this so that #1383 can move forwards?

dprotaso:
/lgtm

Any further changes can be done with subsequent PRs

PR #1479, opened 2024-01-02
Add new approvers to eventing-natss

dan-j:
@astelmashenko and I have been heavily involved in the improvement of the JetStream support in eventing-natss and would like to become approvers of the project. We're already community members so I believe this is the only change we need to make?

At the moment the only approver is @zhaojizhuang, if you would like to provide your 👍 or some feedback on anything better we can do?

Thanks


pierDipi (approved):
/lgtm
/approve

psschwei (approved):

PR #1486, opened 2024-01-11
Chainguard taking ownership of eventing-rabbitmq

dprotaso:
As a results of some restructuring due to the VMware/Broadcom acquisition we're unable to maintain this integration.

I've chatted with Chainguard and they use this integration in production so I'm transitioning approver responsibilities to them.

If anyone else would also like approver status I'd imagine the help would be welcome.

### DIFF
--- a/peribolos/knative-extensions-OWNERS_ALIASES
+++ b/peribolos/knative-extensions-OWNERS_ALIASES
@@ -74,6 +74,7 @@ aliases:
   - lberk
   - matzew
   eventing-rabbitmq-approvers:
+  - gabo1208
   - tcnghia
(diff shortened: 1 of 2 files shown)


dprotaso:
/assign @tcnghia 

/hold there's been some changes in `knative/eventing` that broke the integration (https://github.com/knative-extensions/eventing-rabbitmq/pull/1310)

Will unhold when that's fixed.

dprotaso:
cc @gabo1208

gabo1208:
I could be left as an approver if that's ok

gabo1208 (approved):

PR #1489, opened 2024-01-16
Create folder and files for 2023-2024 end user seat

nainaz:
# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/elections/2023-SC-EU/README.md
+++ b/elections/2023-SC-EU/README.md
@@ -1,0 +1,64 @@
+---
+title: "Knative Steering Committee end user Seat"
+linkTitle: "Steering Committee end user Seat"
+weight: 30
+type: "docs"
+---
+
+# 2023 Knative Steering Committee end user Seat
+
+## Purpose
+
+This document describes the process for filling the end user Representative
+for Knative Steering Committee. The newly appointed representative will serve a
+one (1) year term.
+
+## Background
+
+In 2021 we introduced one end user seat to bring more diversity into the SC
+representation which is to be elected by the SC once the SC election was
+completed.
+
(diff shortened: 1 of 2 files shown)


elections/2023-SC-EU/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Schedule
+
+| Date         | Event                    |
+| ------------ | ------------------------ |
+| January 16  | Announce the intent to appoint end user representative |
+| February 2   | All candidate nominations due by 0000 UTC (4pm Pacific) |
+| February 7   | Announcement of selection |
+
+The one year term would be from February 12, 2024 to February 11 , 2025
```
evankanderson, line 35: I wonder if we want this to continue until the next SC starts the election, rather than a fixed end date.

evankanderson (approved):
/approve

elections/2023-SC-EU/README.md
```diff
@@ -0,0 +1,64 @@
+---
+title: "Knative Steering Committee end user Seat"
+linkTitle: "Steering Committee end user Seat"
+weight: 30
+type: "docs"
+---
+
+# 2023 Knative Steering Committee end user Seat
```
aliok, line 8: I know previous year has been named as 2022 EU seat, but maybe we should stop it and just say 2024 EU seat?

elections/2023-SC-EU/README.md
nainaz: my only concern is how would the end user know about how much commitment are we asking for?

PR #1522, opened 2024-02-28
Use slack ids rather than handles for toc review notifications

psschwei:
# Changes

tldr: Updates TOC review notifications to use Slack ID rather than Handle.

We have a bot that sends slack notifications to WG leads for the TOC review presentations: https://github.com/knative/community/blob/main/.github/workflows/toc-review-reminder.yaml

However, in order for individuals to get notified via Slack, we need to use their Slack ID rather than their handle (as described in [this comment](https://github.com/rtCamp/action-slack-notify/issues/93#issuecomment-854547992)). So this PR makes that update.


dprotaso:
cc @krsna-m 

/lgtm
/approve

PR #1524, opened 2024-03-01
New private Google group for UX WG leads

aliok:
# Changes

- I want to create a new mail alias for the UX WG leads. They're gonna sign up to some systems (Figma, etc) and they want to share their username and password for consolidation.
- So, obviously, this group needs to be private.

### DIFF
--- a/groups/wg-ux/groups.yaml
+++ b/groups/wg-ux/groups.yaml
@@ -1,0 +1,17 @@
+# UX WG
+groups:
+  - email-id: ux@knative.team
+    name: ux
+    description: |-
+      UX WG email
+    settings:
+      WhoCanPostMessage: "ANYONE_CAN_POST"
+      ReconcileMembers: "true"
+    owners:
+      - zainabhusain@ocadu.ca
+      - mariana.mejia@ocadu.ca
+      - leoli@redhat.com
+      - cmurray@redhat.com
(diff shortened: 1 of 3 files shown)


groups/wg-ux/groups.yaml
aliok, line 1: I couldn't find if there's a setting for "private". But I want this group to be private. See PR description above.

groups/wg-ux/groups.yaml
```diff
@@ -0,0 +1,17 @@
...
+    description: |-
+      UX WG email
+    settings:
+      WhoCanPostMessage: "ANYONE_CAN_POST"
+      ReconcileMembers: "true"
+    owners:
+      - zainabhusain@ocadu.ca
+      - mariana.mejia@ocadu.ca
+      - leoli@redhat.com
+      - cmurray@redhat.com
```
aliok, line 14: /hold

@mmejia02 @zainabhusain227 @Leo6Leo @Cali0707 

Are these email addresses good?

Cali0707: These seem to match the emails used on the calendar invite for the meetings so I think it's correct

upodroid (approved):
This configuration creates a private group and membership changes need to be done via  Git.

PR #1521, opened 2024-02-28
Add nrrso to SC

aliok:
# Changes

- Add nrrso to SC
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

PR #1529, opened 2024-03-13
Add Norris to steering / GitHub Org

evankanderson:
# Changes

- Add Norris as the new end-user seat representative to the GitHub org and to steering lists
- Move puerco to Knative membership (Thanks!)
- Note that we do not yet add @nrrso to GitHub groups within the org because he'll need to accept the org invite first to avoid errors.


evankanderson:
/close

Prefer #1521

PR #1534, opened 2024-03-18
Add folks from RabbitMQ Team

ikavgo:
# Changes

/kind enchancement

**Release Note**

```release-note
Add Michal Kuratczyk and Aitor Pérez Cedres from RabbitMQ Team
```

cc @dprotaso @mkuratczyk

### DIFF
--- a/peribolos/knative-extensions.yaml
+++ b/peribolos/knative-extensions.yaml
@@ -97,6 +97,7 @@ orgs:
     - matzew
     - maximilien
     - mgencur
+    - mkuratczyk
     - mmejia02
     - mvinkler
     - nader-ziada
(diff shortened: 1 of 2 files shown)


peribolos/knative-extensions.yaml
```diff
@@ -1025,8 +1025,10 @@ orgs:
           eventing-rabbitmq: write
         members:
         - ikavgo
+        - mkuratczyk
```
dprotaso, line 1028: Let's first add folks to the member list, have then accept their invites. Then after add folks to the rabbitmq team. 

This just avoids the churn with our machinery that updates the OWNER files across the repos.

ikavgo:
/lgtm :-)

@dprotaso I think it's mergeble now

PR #1536, opened 2024-03-25
Request from Ali Ok to remove eric-sap and travis-minke-sap

eric-sap:
# Changes

Removing inactive members eric-sap and travis-minke-sap

### DIFF
--- a/peribolos/knative-extensions.yaml
+++ b/peribolos/knative-extensions.yaml
@@ -638,7 +638,7 @@ orgs:
           Eventing Kafka WG Leads:
             privacy: closed
             description: The Working Group leads for Eventing Kafka
-            members:
+            members: []
       Function Runtime Writers:
         description: Grants write access to knative-extensions/func-go
         privacy: closed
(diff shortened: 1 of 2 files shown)


dprotaso:
You'll need to run the `./hack/update-codegen.sh`

peribolos/knative-extensions.yaml
```diff
@@ -641,7 +639,6 @@ orgs:
             privacy: closed
             description: The Working Group leads for Eventing Kafka
             members:
```
dprotaso, line 641: @pierDipi we should probably consolidate the eventing kafka teams - since this is empty

peribolos/knative-extensions.yaml
```diff
@@ -641,7 +639,6 @@ orgs:
...
-            - travis-minke-sap
```
dprotaso, line 644: @eric-sap you might need to replace this an empty list `[]`

dprotaso:
/lgtm
/approve

thanks 🎉 - you're welcome back anytime

PR #1533, opened 2024-03-15
remove nrrso until he accepts GitHub org invites

dprotaso:
This unblocks the OWNER files being propagated until nrrso accepts the GitHub invites.

Once he accepts we can add him back to the steering team

PR #1537, opened 2024-03-25
Drop extra files that were committed in #1533

dprotaso:
I added some files I didn't expect to add. The active user list is
probably over 2 years old (that's the last time I pruned the list).

/assign @upodroid


upodroid (approved):
LGTM

PR #1540, opened 2024-03-26
Replace steven0711dong with pierDipi in MT kafka source group

aliok:
# Changes

- Knative SC is removing inactive approvers and WG leads as it is a bad project governing practice.
- Replace steven0711dong with pierDipi in MT kafka source group.
- That group doesn't have any folks other than @steven0711dong, so, I've added eventing lead @pierDipi there
- I was unable to contact @steven0711dong over CNCF Slack or email.

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/peribolos/knative-extensions-OWNERS_ALIASES
+++ b/peribolos/knative-extensions-OWNERS_ALIASES
@@ -52,7 +52,6 @@ aliases:
   - lberk
   - matzew
   - pierDipi
-  - steven0711dong
   eventing-kafka-broker-approvers:
   - aliok
   - matzew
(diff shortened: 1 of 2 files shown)


aliok:
cc @pierDipi

peribolos/knative-extensions-OWNERS_ALIASES
```diff
@@ -61,7 +60,7 @@ aliases:
   - Leo6Leo
   - cali0707
   eventing-kafka-mtsource-approvers:
-  - steven0711dong
+  - pierDipi
```
pierDipi, line 63: The repository is deperecated and eventing-writers is always an owener so we don't really need anyone in this group

aliok:
/close

I created an additional PR for removing obsolete groups: https://github.com/knative/community/pull/1547

That should fix this issue.

PR #1547, opened 2024-03-26
Remove emeritus WGs from Peribolos

aliok:
# Changes

- Remove emeritus WGs from Peribolos
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/peribolos/knative-OWNERS_ALIASES
+++ b/peribolos/knative-OWNERS_ALIASES
@@ -14,37 +14,27 @@ aliases:
   - navidshaikh
   - rhuss
   - vyasgun
-  conformance-task-force-leads:
-  - salaboy
-  conformance-writers:
-  - salaboy
   docs-reviewers:
   - nainaz
   - pmbanugo
   - retocode
   - skonto
   - snneji
-  docs-wg-leads:
-  - snneji
   docs-writers:
   - csantanapr
   - retocode
   - skonto
-  - snneji
   eventing-reviewers:
   - Leo6Leo
   - aslom
   - cali0707
   - creydr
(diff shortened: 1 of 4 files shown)


aliok:
@cardil @upodroid 

I am not 100% sure if the removed things are referenced in any repos. Can they be used in the OWNERS files in repositories? Or, do we manage those files based on the files here and this PR will be just fine?

cardil (approved):

cardil:
> @aliok: I am not 100% sure if the removed things are referenced in any repos. Can they be used in the OWNERS files in repositories? Or, do we manage those files based on the files here and this PR will be just fine?

I think only OWNER_ALIASES is synched. The OWNERS file needs to be adjusted if those removed teams are used. IDK, if it will break Prow review suggestions, but I doubt it.

aliok:
> The OWNERS file needs to be adjusted if those removed teams are used. IDK, if it will break Prow review suggestions, but I doubt it.

Do I need to go and check each repo to find out where these deleted teams are used? Or will it be "only" review suggestions that might be broken. If it is the latter, I would want to merge this PR and send an email to knative-dev asking folks to check their own repositories.

upodroid (approved):
We aren't using these WGs in OWNERS file.

https://github.com/search?type=code&q=org%3Aknative+OR+org%3Aknative-extensions+%28conformance-task-force-leads+OR+docs-wg-leads+OR+eventing-triage+OR+conformance-writers%29+AND+NOT+path%3AOWNERS_ALIASES+

/hold cancel

PR #1548, opened 2024-03-29
Add rhuss and dsimansk as approvers to all kn plugins

aliok:
# Changes

- Add rhuss and dsimansk as approvers to all kn plugins
- This is necessary to remove some inactive approvers later on


aliok:
cc @rhuss @dsimansk

psschwei:
Are Roland / David covered as members of client-writers? For example, on quickstart we've explicitly listed them as one of the owners: https://github.com/knative-extensions/kn-plugin-quickstart/blob/main/OWNERS#L4

dprotaso:
Yeah let's update the OWNER files in the various repos to add `client-writers`

psschwei:
I did a quick check of the kn-plugin repos and all looked to have client-writers in the owners file. So I think we're all good here.

davidhadas:
Need rebase

aliok:
> I did a quick check of the kn-plugin repos and all looked to have client-writers in the owners file. So I think we're all good here.

Thanks for checking. I am closing this one then.

aliok:
Maybe I misunderstood...

Can you have a look at https://github.com/knative/community/blob/b056e6e06a8d413a1c40789d86199e592faccd80/peribolos/knative-extensions-OWNERS_ALIASES#L127-L129

Both names there are inactive and to be removed. Does that mean that leaving the group empty will work since rhuss and dsimansk are in the client-writers groups in all these repositories?

cc @dprotaso

psschwei:
> Does that mean that leaving the group empty will work since rhuss and dsimansk are in the client-writers groups in all these repositories?

Yes, that's right

PR #1551, opened 2024-04-17
paul stepping down as serving approver

psschwei:
# Changes

I haven't been actively contributing to serving for a while now ( :frowning_face: ), so I should probably go ahead and remove myself as a serving approver.

/assign @dprotaso


dprotaso:
😢 
/lgtm
/approve

PR #1568, opened 2024-05-03
Update TOC review schedule

dsimansk:
- Update TOC review schedule

/cc @knative/technical-oversight-committee


psschwei (commented):
/lgtm

PR #1569, opened 2024-05-06
Revoke release leads permissions since 1.14 is out

izabelacg:
# Changes

- Revoke release leads permissions since 1.14 is out


Cali0707 (commented):
/lgtm from me
/cc @knative/technical-oversight-committee

dsimansk:
I believe you need to run `./hack/update-codegen.sh` to regen owners files based on peribolos config. Ci checks should catch it, let' see.

PR #1605, opened 2024-08-22
Dave is now independent

dprotaso:


dprotaso:
cc @knative/steering-committee needs to approve this

evankanderson (approved):

PR #1587, opened 2024-07-01
Combine Steering and TO Committees

evankanderson:
# Changes

- Unify the Steering Committee and TOC.
- Adjust the _future_ membership to consist of 4 elected seats, 1 end-user seat, and up to 2 selected seats to balance the set of skills on the committee.

This is based on https://github.com/knative/community/issues/1549#issuecomment-2147974257, but I've adjusted the Steering-appointed ("selected") seats to be optional and 1-year terms rather than 2-year terms with staggered appointment.  I don't feel strongly about this, but it seemed to provide more flexibility.

Relates to #1549

PR #1613, opened 2024-09-23
Edit Github admin teams and drop TOC review reminder

psschwei:
# Changes

As TOC was folded into Steering in #1587 , moving the former TOC members into the Steering github team. 

We also no longer need to post reminders about the TOC review meetings.

/cc @knative/steering-committee @knative/technical-oversight-committee

### DIFF
--- a/.github/workflows/toc-review-reminder.yaml
+++ b/.github/workflows/toc-review-reminder.yaml
@@ -12,48 +12,48 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: 'TOC review reminder'
+# name: 'TOC review reminder'
 
-on:
-  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


aliok (commented):
/lgtm
/approve
/hold

Holding for productivity team review, if necessary.

.github/workflows/toc-review-reminder.yaml
```diff
@@ -1,59 +0,0 @@
-# Copyright 2024 The Knative Authors.
```
dprotaso, line 1: Maybe disable the  job - might be worth discussing if steering wants to continue this practice

psschwei:
Think we can drop the hold here, but will wait a day or so in case @knative/productivity-reviewers think otherwise

Issue #1558, opened 2024-04-24
New Repo: `func-python`


matzew:
Use this issue type to request a new repo in `knative-extensions` (or
`knative`, which may require additional discussion).

## Repo information

Org: `knative-extensions`

Repo: `func-python`

Purpose (Description):

```
The repository contains the function invocation and runtime framework for Knative Python functions.
```

Sponsoring WG:
Knative Functions/Clients

## Actions to fulfill

This area is used to track the [repo creation process](https://github.com/knative/community/blob/main/mechanics/CREATING-AN-EXTENSIONS-REPO.md).
The _requestor_ and _sponsoring WG lead_ should perform the steps listed below and cross out the checkmarks when done.
The TOC is involved only in the **TOC Gate** steps.

- [x] Add this issue to the [TOC project board](https://github.com/orgs/knative/projects/43) for review. You are responsible for moving your entry on the board to "Needs Discussion" or "In Progress" as you move forward in this checklist.

_You may not be able to use the Projects quick menu on this page. In that case, go to the project board and use the **Add cards** interface._

- [x] Send a PR adding entries for this repo in [`/peribolos/knative-extensions.yaml`](https://github.com/knative/community/blob/main/peribolos/knative-extensions.yaml). Please mind the alphabetical order when adding to a list.
  - [x] Add the repository and a description.
  - [x] Grant `Knative Admin` the `admin` privilege.
  - [x] Grant the sponsoring WG the `write` privilege.

**TOC Gate**: _Once the TOC has approved the above, it will merge and Peribolos will create an empty repository._

- [x] Have a lead from the sponsoring WG bootstrap the Git repository by using an 
  appropriate "template" repository ([basic](https://github.com/knative-extensions/wg-repository),
  [sample-controller](https://github.com/knative-extensions/sample-controller),
  [sample-source](https://github.com/knative-extensions/sample-source)).

  Follow the instructions from [Creating a repository from a template](https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-repository-from-a-template)

- [x] Send a PR adding the repo to [knobots](https://github.com/knative-extensions/knobots). Then run the community files update action to propagate new OWNER_ALIASES.

- [x] Add your GitHub Team to the `OWNERS` file for your repo.

- [x] [Set up](https://github.com/knative/test-infra/blob/main/guides/prow_knative_setup.md#setting-up-prow-for-a-new-repo-reviewers-assignment-and-auto-merge) prow for a new repo

[...]

PR #1619, opened 2024-10-01
Clarify for golang only doing hack

matzew:
# Changes

- based on [discussion](https://github.com/knative/community/issues/1558#issuecomment-2378442775) it feels that `hack` setup is useful for golang only - hence updating the template

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


dprotaso:
technically it's for anything - but the default ones are golang

you just need to include a bash function in your scripts to override it 

eg. https://github.com/knative/hack/blob/fc6a8452af6d002ab7fb2818520a6f3d105ee1a0/presubmit-tests.sh#L179

dprotaso:
Closing as per above comment

PR #1621, opened 2024-10-10
Use python 3.11 for peribolos check

dsimansk:
Related to https://github.com/knative/infra/issues/530

/cc @skonto @ReToCode @upodroid

### DIFF
--- a/.github/workflows/peribolos-schema-validate.yaml
+++ b/.github/workflows/peribolos-schema-validate.yaml
@@ -29,7 +29,10 @@ jobs:
       with:
         fetch-depth: 0
         lfs: true
-    - run: pip install strictyaml --break-system-packages
+    - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"    
+    - run: pip install strictyaml
     - name: Validate knative.yaml
       run: python3 peribolos/validate-schema.py peribolos/knative.yaml
     - name: Validate knative-extensions.yaml


peribolos/ci
```diff
@@ -0,0 +1 @@
+Thu Oct 10 15:01:07 CEST 2024
```
dsimansk, line 1: Forcing an action run, should be removed before merge.

/hold

.github/workflows/peribolos-schema-validate.yaml
```diff
@@ -29,7 +29,10 @@ jobs:
       with:
         fetch-depth: 0
         lfs: true
-    - run: pip install strictyaml --break-system-packages
+    - uses: actions/setup-python@v5
+      with:
+        python-version: "3.11"    
+    - run: pip install strictyaml
```
dsimansk, line 35: It seems we can go with `--break-system-packages` or pin Python to a version that works without the flag. I don't know if there are any drawbacks with `--break-system-packages`, but it's one liner in a check we always can revisit.

dsimansk: `pipx` doesn't work for libraries:
```
➜  community git:(pr/pipx-fix) pipx install strictyaml

No apps associated with package strictyaml or its dependencies. If you are attempting to install a library, pipx should not be used. Consider using pip or a similar tool instead.
```

peribolos/ci
```diff
@@ -0,0 +1 @@
+Thu Oct 10 15:10:16 CEST 2024
```
dsimansk, line 1: Forcing an action run, should be removed in final version.

ReToCode:
@dsimansk we went with the other approach, right?

ReToCode (approved):
/lgtm
/approve

PR #1623, opened 2024-10-17
Replacing some more TOC references with KSC ones

psschwei:
Updating some references now that we've merged TOC and KSC.

Would be good to get eyes from Productivity to validate that it's ok to remove the TOC group

/assign @knative/steering-committee @knative/productivity-writers


evankanderson (approved):
/lgtm

PR #1624, opened 2024-10-23
Remove me from rabbitmq-eventing maintainers

Zerpet:
# Changes

- 🗑️ Remove me from the project

/kind removal

Unfortunately, I have to withdraw from the project and focus in other project.

### DIFF
--- a/peribolos/knative.yaml
+++ b/peribolos/knative.yaml
@@ -152,7 +152,6 @@ orgs:
     - yuvalif
     - yuzisun
     - zainabhusain227
-    - Zerpet
     - zhanggbj
     - zhaojizhuang
     - ZhiminXiang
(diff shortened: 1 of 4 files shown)


dsimansk:
@Zerpet sad news. 

To drop yourself from approvers, you will need to remote this entry: https://github.com/knative/community/blob/main/peribolos/knative-extensions.yaml#L1041

and then run `./hack/update-codege.sh` to regen owners files.

peribolos/knative.yaml
```diff
@@ -152,7 +152,6 @@ orgs:
     - yuvalif
     - yuzisun
     - zainabhusain227
-    - Zerpet
```
dsimansk, line 155: This line will effectively remove you from `Knative` gh org, if that's also desired. Then, you want to do the same on knative-extensions.yaml. But that's solely on your decision AFAIK.

Zerpet: Yeah, this is ok. I don't think I'll get time to contribute to this community any time soon 😔

PR #1639, opened 2024-11-26
drop HighCohesion as an adopter

dprotaso:
Reached out and they don't use Knative anymore


dprotaso:

psschwei (commented):
/lgtm

PR #1643, opened 2024-11-28
Remove owners field in 2024 SC election config

creydr:
Copy & past issue.
Owners field is not needed in election config, as it does not exist in candidate template

/cc @jberkus


dsimansk (approved):
/approve
/lgtm

PR #1645, opened 2024-12-02
Extend nominations deadline for SC election

dsimansk:
# Changes

- Extend nominations deadline

PTAL take a look if we want to move the whole schedule accordingly. Or we can vote on that in a follow-up as well. 

### DIFF
--- a/elections/2024-SC/README.md
+++ b/elections/2024-SC/README.md
@@ -34,11 +34,14 @@ to check if you are eligible by logging into [Elekto]. See Voting Process below 
 | Date         | Event                    |
 | ------------ | ------------------------ |
 | November 20  | Announcement of Election, call for nominations, exceptions |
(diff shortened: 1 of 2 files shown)


jberkus:
We'll need to update election.yaml file as well, with the new dates.

psschwei (commented):
+1

evankanderson (commented):
:+1:

This seems good to me.  I was out last week, but I think I know at least one more candidate who may apply.

dsimansk:
> We'll need to update election.yaml file as well, with the new dates.

Thanks, updated!

davidhadas:
+1

dprotaso:
+1

dsimansk:
+1 to explicitly confirm 5 SC votes

jberkus:
Config file looks good.

aliok:
+1

PR #1648, opened 2024-12-10
TATA Comm uses Knative in production

dprotaso:
They use Knative to build out their internal services instead of relying on controllers/operators


dprotaso:

PR #1650, opened 2024-12-12
Add Cloudera as an adopter

dprotaso:
They blogged about it here: https://blog.cloudera.com/cloudera-ai-inference-service-enables-easy-integration-and-deployment-of-genai-into-your-production-environments/


dprotaso:

PR #1651, opened 2024-12-12
Add 2024 SC Member Election Results

creydr:
As the [announcement](https://groups.google.com/g/knative-users/c/kMiMsJCabRk) was sent, adding the results here too


dprotaso:
cc @matzew

PR #1664, opened 2025-01-28
Update release schedule for next cycle

dsimansk:
# Changes

- Update release schedule for next cycle

/cc @knative/knative-release-leads @knative/steering-committee


mechanics/RELEASE-SCHEDULE.md
```diff
@@ -22,21 +22,22 @@ With that said, it can be useful to have a list of when future releases will hap
 
 | Release | Date       | EOL        | Min K8s Version | Notes |
 |---------|------------|------------|-----------------| ----- |
-| 1.17    | 2025-01-28 | 2025-07-28 | 1.30            |       |
 | 1.18    | 2025-04-22 | 2025-10-28 | 1.31            |       |
 | 1.19    | 2025-07-22 | 2026-01-28 | 1.32            |       |
+| 1.20    | 2025-10-21 | 2026-04-21 | 1.33            |       |
```
dprotaso, line 27: Unsure if k8s changes their patch release dates but typically we support the min version that isn't in maintenance mode/EOL - give or take a week.

- 1.18 is releases a week before k8s 1.30 is in maintenance mode so 1.31 makes sense
- 1.19 targets 1.32 - but 1.31 goes into maintenance mode a _month_ after release so that should be 1.31
- 1.20 targets 1.33 but 1.32 goes into maintenance mode _two months_ after release so that should be 1.32

so 
```suggestion
| 1.18    | 2025-04-22 | 2025-10-28 | 1.31            |       |
| 1.19    | 2025-07-22 | 2026-01-28 | 1.32            |       |
| 1.20    | 2025-10-21 | 2026-04-21 | 1.32            |       |
```

dprotaso: See  https://kubernetes.io/releases/patch-releases/

dsimansk: Thanks, sure makes sense. I haven't checked k8s schedule TBH. :)

PR #1681, opened 2025-04-23
Add gauron99 as a regular member of knative-extensions

dsimansk:
# Changes

- Add gauron99 as a regular member of knative-extensions

This should fix the issue with recurring new invites. You wasn't added to extensions org config as a member.
/cc @gauron99


gauron99 (approved):

gauron99:
aha thats what it was, thanks!

PR #1688, opened 2025-05-28
Clarify policies for removal of steering committee members

### LINKED ISSUE #1687: Governance - Add requirements for remaining a KSC member / removing a KSC member
From doing a governance review of Knative.

The governance docs are very complete. The part of the KSC lifecycle that isn't documented is: 
- Requirements for remaining a KSC member
- How a member would be removed if they didn't meet the requirements

A simple approach may be to implement something similar to Dapr, in STEERING-COMMITTEE.md:

**<suggestion>**
 "KSC members MUST remain active. If they are unresponsive for >3 months, they will be automatically removed unless a super-majority of the other KSC members agrees to extend the period to be greater than 3 months."
**</suggestion>**

I'm raising this as an issue. As you work with TOC, together you can decide whether this is important enough to implement

evankanderson:
# Changes

- Documents the expectation that steering committee members remain active.
- Documents the process for removing a steering committee member, either due to inactivity, or for cause.
- Clears up an ambiguous section which defined a _second_ way of filling a KSC vacancy.  (Oops!)

/kind governance
Fixes #1687

Note that this is purely to address feedback from the governance review, and mostly just clarifies rules that we'd make up on the spot (using the charter approval process) if we needed them.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,0 +1,22 @@
+---
+title: Knative Governance Model
+linkTitle: Governance
+weight: 20
+type: "docs"
+aliases:
+  - /contributing/governance
+---
+
+Knative has used a number of different governance models in the past. The
+_current_ governance of the Knative project is managed by the [Knative Steering
+Committee](./STEERING-COMMITTEE.md), which delegates much of the technical work
+to different [working groups](./working-groups/WORKING-GROUPS.md), which are
+focused on specific components of Knative.
+
+All participants in the Knative project are contributors; contributors who
+reach a minimum level of interactions ([currently 25 interactions in 12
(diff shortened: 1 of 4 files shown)


evankanderson:
/hold

This is a charter amendment, so will need some vote.

evankanderson:
For reference, for non-Steering roles, we offer a fairly generous 12-month inactivity window:

https://github.com/knative/community/blob/main/ROLES.md#inactivity

I think a shorter window for Steering makes sense given the requirements for quorum votes which none of the other roles have.

dprotaso:
+1 from me

we have quorum
/hold cancel


cc @matzew

PR #1701, opened 2025-08-07
drop arsenator from steering team until he accepts the org invite

dprotaso:


dprotaso:
looks like he accepted the invite

PR #1731, opened 2025-11-04
fix: voters have correct case sensitivity

Cali0707:
# Changes

- Voter github handles are case sensitive in elekto, so our list now has the correct capitalization


geekygirldawn (approved):
/lgtm

